PR #3, opened 2026-04-27
JSON Serialization

Quant-TheodoreFelix:
This PR implements the following main changes:

* Added the `serde` dependency
* Defined the JSON schema and implemented serialization (written stably in `core.rs`)
* Added the following subcommands:
  * `ast-outline json [paths...]` (default outline view)
  * `ast-outline json outline <paths> --no-private --no-fields --no-docs --no-attrs --glob --compact`
  * `ast-outline json digest <paths> --include-private --include-fields`
  * `ast-outline json show <path> <symbol>...`
  * `ast-outline json implements <target> <paths> --direct`

For the subcommands, I added various flags to allow flexible output. Since this feature was created by adding or extending existing commands under the “JSON serialization” design, I think it may differ slightly from the main design direction. If you find them unnecessary, feel free to remove or modify them.

I also performed validation testing. Using sample code across all supported languages (Rust, Python, TS, JS, C#, Go, Java, Kotlin, Scala, Markdown), I ran validation in Python and confirmed everything works correctly — trees, nesting, line numbers, visibility, filters, and cross-language implementations all function as expected.

I left comments throughout the code and changes! You can simply replace the regex `// Q. T. Felix - [a-z]{3,5}` with a space in the codebase 😂

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 
 **/*.rs.bk
 Cargo.lock
+
+/.idea
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,8 @@ ast-grep-language = "0.42.1"
 tree-sitter = "0.26.3"
 clap = { version = "4.5", features = ["derive"] }
 ignore = "0.4.22"
-rayon = "1.10"
+rayon = "1.10" # yoo what is this
 tree-sitter-md = "0.5.3"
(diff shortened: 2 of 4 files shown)


aeroxy:
Awesome man! Let me think about if adding a `json` cmd is the best approach. Thank you so much!

PR #9, opened 2026-04-28
Some more clean ups

vkarasen:
1. moved to 2024 rust edition for newer language features like let-chains
2. fixed all clippy warnings, `nix flake check .` doesn't show errors anymore
3. reorganized the flake
    * moved from flake-utils to flake parts
    * removed cargo-fmt and toml-fmt checks -> don't want to incur on formatting

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -30,6 +30,32 @@ for that file is partial — read the source directly for the affected region.
 
 `ast-outline help` for flags and rare options.
 
+## JSON / machine-readable output
+
+Add `--json` to any command for stable, structured output instead of
+formatted text — useful when you are parsing the result programmatically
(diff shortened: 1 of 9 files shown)


aeroxy:
Thanks man! I appreciate your drive to modernize the project, but I'm going to push back on this specific change. edition = "2024" would require Rust 1.85, and that's simply too new. Our potential users in finance and other regulated industries are almost certainly on older, validated toolchains. Let's hold off on the edition upgrade for now, as it would prematurely limit our user base.

vkarasen:
ah sure, I didn't think you'd care. If you're set on 2021 I'll change the commit to only fix the linter warnings that come up under it, really I just want the checks to pass ;)

PR #22, opened 2026-06-02
Fix: Glob support for filepath in cargo.toml

adamhumphrey44:
In src/surface/manifest.rs, parse_cargo_toml() reads [workspace].members with _parse_string_array(value) and stores entries like "crates/*" literally, without expanding Cargo workspace globs.

In src/surface/entry_point.rs, discover_rust() iterates manifest.workspace_members, builds member_root = manifest.manifest_dir.join(m), and calls discover_rust(&member_root), so it tries to inspect a literal path like ./crates/* instead of the real crate directories.

When no workspace members resolve, discover_rust() falls through to _resolve_rust_root(), fails to find root-level src/lib.rs, src/main.rs, [lib].path, or [[bin]].path, and emits Cargo.toml found but no lib.rs/main.rs and no [lib].path/[[bin]].path entry.

Patched with a glob fallback.

## Summary by CodeRabbit

* **Improvements**
  * Workspace member paths now support glob patterns for richer project discovery.
  * Better path expansion and resolution for files and workspaces, reducing missed crates.
  * Cycle detection in Rust project discovery to avoid infinite recursion.
  * Workspace exclusion support to respect workspace ignore lists.
  * Minor UX/message refinements for clearer diagnostics during discovery and parsing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -92,6 +92,7 @@ dependencies = [
  "colored",
  "dirs",
  "fs2",
+ "glob",
  "ignore",
  "memmap2",
  "rayon",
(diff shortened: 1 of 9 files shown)


aeroxy:
@codex review

aeroxy:
@jose44github good job! thx

PR #26, opened 2026-06-15
fix: add markdown files to crane source filter

vkarasen:
Flake stopped working since commit
e9cfb2fc0cdc0dd72df505c5babc93fe4196ba1e started to include the prompt directly from skill file.

This is because crane (and nix in general) will only copy certain subsets of files to the build sandbox.

This commit adds markdown files to the filter as well


aeroxy:
Thank you!

Issue #37, opened 2026-07-28
Merge `digest` into `map`: one command, orthogonal flags, named presets


vlsi:
## Summary

`map` and `digest` are already one command internally — they emit byte-identical JSON — but the CLI splits them into two, with disjoint flag sets of opposite polarity. The split forces a caller to choose a detail level and a scoping mechanism together, when the two are independent: there is no way to cap members on `map`, and no way to filter by glob on `digest`. Please make it one command with orthogonal flags, and keep `digest` as a named preset.

Measured with `ast-bro 3.0.0` (Homebrew `aeroxy/tap`, bottle 64), macOS 26.5.1 arm64, on the 30-file Java package `org.postgresql.api.codec` in [pgjdbc](https://github.com/pgjdbc/pgjdbc).

## They already share one model

```console
$ sb map    pgjdbc/src/main/java/org/postgresql/api/codec/CodecLookup.java --json --compact > /tmp/m.json
$ sb digest pgjdbc/src/main/java/org/postgresql/api/codec/CodecLookup.java --json --compact > /tmp/d.json
$ cmp /tmp/m.json /tmp/d.json && echo identical
identical
```

5,229 bytes, same `ast-bro.map.v1` envelope, same declarations, same `signature` on every one. `--json` erases the distinction entirely: the two subcommands differ only in their text renderer and in which flags they accept.

## The flag sets are disjoint, and opposite in polarity

| | flags |
| --- | --- |
| `map` | `--no-private`, `--no-fields`, `--no-docs`, `--no-attrs`, `--no-lines`, `--glob` |
| `digest` | `--include-private`, `--include-fields`, `--max-members` |

Nothing is shared except `--json` / `--compact`. `map` shows everything and subtracts; `digest` shows nothing and adds. So `--no-fields` and `--include-fields` control the same underlying switch under two names, on two commands, in two directions.

Two consequences for a caller:

**Axes cannot be combined.** Detail level and scope are independent questions, but the CLI ties both to the command name. `sb map <dir> --max-members 8` — a signature-level view with a per-file cap — is not expressible, and neither is `sb digest <dir> --glob '*.java'`.

**Choosing the wrong command is silent.** Both of those calls print the top-level help on stdout and exit 0, so a mistyped flag is indistinguishable from a large successful answer:

```console
$ sb map <file> --max-members 3 ; echo "exit=$?"
Fast, AST-based structural outline for source files
…
exit=0
[...]

Issue #35, opened 2026-07-28
map on a directory: point at digest when the output gets large


vlsi:
## Summary

`map` accepts a directory and answers it literally, so a package-wide question produces an output nobody can use as one piece — and nothing in that output mentions that `digest` answers the same question at a sixth of the size. The natural next move is to pipe it through `head`, which cuts the middle of the package away silently.

Measured on the 30-file Java package `org.postgresql.api.codec` in pgjdbc, `ast-bro 3.0.0` (Homebrew 64), macOS 26.5.1 arm64:

| Command | Output |
| --- | --- |
| `sb map <dir>` | 140,714 B |
| `sb map <dir> --no-docs --no-fields --no-private` | 28,372 B |
| `sb digest <dir>` | 9,684 B |
| `sb digest <dir> --max-members 8` | 8,613 B |

I was reviewing naming consistency across that package. `digest --max-members 8` fits in one read and carries every method name — precisely the input the question needed. I reached for `map` instead, hit 28 KB even with three narrowing flags, cut it with `head -200` and `tail -120`, and four files in the middle never appeared in either half. The report I wrote from it had a hole I could not see.

## Request

Print a hint when `map` is given a directory **and** the output exceeds a threshold (20–30 KB seems about right — small directories should stay quiet). For example, as a trailing line:

```
# hint: this was a directory; `ast-bro digest <dir> --max-members 8` answers the same question in ~8 KB
```

## Please do not silently redirect `map <dir>` to `digest`

The two have different contracts and I rely on the difference: `map` prints signatures with doc comments verbatim, which is why it is the right tool for reviewing one file's API, while `digest` prints names without signatures. Rewriting the command based on the argument's type would make the tool return something other than what was asked for — the same class of surprise as an output that looks like an error when nothing failed. A hint keeps the choice with the caller.

Related: #33 covers a different failure of `map`/`digest` on paths.

aeroxy:
Shipped. `map` on a directory now prints the hint on stderr above a size threshold, and stays quiet below it — verified on `main` (d993b62):

```
src/hook      ->   3,286 B stdout, no hint
src/squeeze   ->   9,025 B stdout, no hint
src/adapters  ->  39,990 B stdout, hint
src/          -> 285,541 B stdout, hint
```

```
# hint: this was a directory (39 KB); `map src/adapters --preset digest --max-members 8`
# answers the same question in a fraction of the size
```

[...]

Issue #34, opened 2026-07-28
hook: substitution reads as a tool failure, and the explanation lands last


vlsi:
## Summary

The Claude Code hook substitutes a file's map for a `Read`, which is exactly what I want it to do. It delivers that substitution as `{"decision":"block", "reason":"<the map>"}`, and the client renders a blocked tool call as a tool **error**. So the model reads "this call failed" first, and the single line saying that nothing failed sits at the very end of a multi-kilobyte payload — past the point where the decision to abandon the result has already been made.

In today's session I asked for `Read` on a 212-line Java file, got the map back inside an error wrapper, did not register it as the answer, and re-ran the same lookup through a different command. The information was there; the framing hid it.

Version: `ast-bro 3.0.0` (Homebrew 64), macOS 26.5.1, Darwin 25.5.0 arm64. Hook config in `~/.claude/settings.json`: PreToolUse matcher `Read` → `ast-bro hook --protocol claude-code --min-lines 200`.

## What the hook returns

```console
$ echo '{"tool_name":"Read","tool_input":{"file_path":".../CodecContext.java"}}' | ast-bro hook --protocol claude-code --min-lines 200
{"decision":"block","reason":"# .../CodecContext.java ([medium], 212 lines, 9022 chars, 1 types, 16 methods)\n … ~9 KB of map …\n\n# ast-bro substituted full file. Re-read with offset/limit, or\n# `ast-bro show <file> <symbol>` for a body.\n"}
```

The client shows that `reason` to the model as the content of a failed tool call.

## Why it matters

A blocked call and a successful substitution call for opposite reactions. The first says "change your approach"; the second says "use this, it is what you asked for in a better form". Encoding the second as the first spends the reader's attention on recovering from a failure that did not happen, and the disambiguating sentence arrives thousands of tokens too late to prevent it.

## Suggestion

1. Lead with the explanation. Something like `# ast-bro substituted the map for Read: nothing failed. Line ranges below; re-read with offset/limit, or ast-bro show <file> <symbol> for a body.` as the **first** line of `reason`, with the map underneath. Keeping the trailing copy is fine.
2. If the claude-code hook protocol offers a channel that does not mark the result as a failure — additional context on an allowed call rather than a block — prefer it. A substitution is not a denial, and the transcript should not say it is.

For reference, the hook correctly leaves alone what it cannot parse: a `.txt` path (including ast-bro's own saved output) returns `{"continue":true}`.

vlsi:
[...]

Issue #33, opened 2026-07-28
map/digest exit 0 with empty output when no paths resolve


vlsi:
## What happens

`ast-bro map` with no paths exits 0 and prints zero bytes. With `--json` it prints a well-formed `{"schema": "ast-bro.map.v1", "files": []}`. A path that does not exist prints a `# note:` line and also exits 0.

| Command | exit | stdout |
| --- | --- | --- |
| `ast-bro map` | 0 | (empty) |
| `ast-bro map --json` | 0 | `{"schema": "ast-bro.map.v1", "files": []}` |
| `ast-bro map /nope/x.java` | 0 | `# note: path not found: /nope/x.java` |
| `ast-bro digest` | 0 | `# no files` |

Measured with `ast-bro 3.0.0` on macOS.

## Why it matters for an AI agent

An agent composes arguments from a previous step, most often a shell substitution:

```bash
ast-bro map $(some-file-finder 'CodecContext*.java')
```

When the finder returns nothing, the substitution collapses and `map` receives no paths. The agent sees an empty result and reads it as an answer about the code: "this package declares nothing." It then proceeds on that false premise instead of fixing the argument list. The JSON form is worse than the text form, because `files: []` is a valid, confident-looking answer.

An agent cannot distinguish "you asked about nothing" from "the answer is nothing" unless the tool says so. Exit code and stderr are the two channels it already checks.

This is not hypothetical: it cost a real session a wasted turn, and the empty output was initially read as a fact about the package.

## Proposed behavior

Adopt one invariant: **no subcommand exits 0 with an empty stdout, and no subcommand reports an empty result when the input itself was empty.**

1. No paths after argument expansion → exit 2, message on stderr, nothing on stdout.
2. Zero of N given paths resolved → exit 2. Some but not all resolved → exit 0, keep the existing `# note:` line on stderr.
3. In `--json` mode, replace the empty payload with an error object on stderr, so a parser never receives a valid-looking empty result for an invalid request.

Suggested message. It names what arrived, why that is an error rather than a result, and what to do — the three things an agent needs to recover without guessing:

```text
error: `map` received no paths (0 arguments after the subcommand).
  An empty argument list is not an empty codebase: nothing was inspected.
  If the list came from a shell substitution such as `ast-bro map $(...)`, the
  substitution produced nothing. Resolve the files first, check the list is non-empty,
  then pass the paths explicitly.
usage: ast-bro map [PATHS...]
```

and for unresolved paths:

```text
[...]

PR #38, opened 2026-07-28
feat: agent-safe CLI contract, resolver correctness, map/digest merge (#31-#37)

aeroxy:
Resolves the seven issues filed against 3.0.0, plus review follow-ups:

- #31 callers under-reports on receiver shadowing: pass A now respects explicit receivers (self-like keywords only; type-qualified calls bind to the scoped qn), self-like binding prefers the caller's own scope, and edges the resolver could not attribute are surfaced in callers output (text section + JSON unattributed/_total/_truncated/_hidden) instead of silently dropped. PHP scope keywords are case-folded.
- #33/#36 one error contract: stdout carries results only; exit 0 = query ran (empty answers included), 2 = query could not run as asked, 1 = internal failure; rejections print to stderr and emit an ast-bro.error.v1 envelope under --json (new src/cli_error.rs). Unknown flags exit 2 with a cross-subcommand hint instead of help-on-stdout. Empty/unresolved path lists are rejected before walking.
- #32 truncation is never silent: callers/callees/reverse-deps/impact report true totals when --limit trims display, JSON carries total/truncated (+ frontier_truncated for depth-stopped walks, noted on stderr only when --depth was raised), map --max-members prints +N more and JSON dropped_members agrees with the text renderer.
- #37 map and digest are one command: --detail names|signatures|full, one-polarity visibility flags, scope flags on both; digest is an alias for map --preset digest with explicit flags overriding the preset.
- #35 map on a large directory hints at the digest preset (stderr, never a silent redirect).
- #34 the Read-hook substitution leads with "nothing failed" so the map is read as the answer, not an error.

MCP mirrors the CLI: trace rejections render a valid trace.v1 doc with found:false, callers folds hidden-unattributed counts into the response body, and the unattributed section shares the --limit budget everywhere.

## Summary by CodeRabbit

* **New Features**
  * Added consistent `map` and `digest` options, presets, detail levels, member limits, projections, and truncation reporting.
  * Added result limits and exact totals for callers, callees, impact, and reverse-dependency searches.
  * Added richer unresolved, ambiguous, and unattributed call-site reporting.
  * Added support for indexing TOML and PowerShell files.
  * Added broader embedding compatibility, including `f16` data.

* **Bug Fixes**
  * Standardized validation errors, exit codes, stdout/stderr behavior, and machine-readable JSON error responses.
  * Improved receiver-aware call resolution and incremental analysis consistency.
  * Fixed case-insensitive scope keyword matching in PHP call resolution.
  * Refined visibility classification for trait members and module-scoped items.

* **Documentation**
  * Updated CLI, call-analysis, search, filtering, and architecture documentation.


aeroxy:
@coderabbitai full review

aeroxy:
@coderabbitai full review

aeroxy:
@coderabbitai full review

src/calls/traverse.rs
```diff
@@ -61,21 +88,118 @@ pub fn callers<F: Fn(&CallEdge) -> bool>(
...
+/// a rename as cheaper than it is (issue #31).
+pub fn unattributed_callers(graph: &CallGraph, targets: &[Qn]) -> Vec<CallEdge> {
+    use crate::calls::graph::CallTarget;
+    let target_set: HashSet<&Qn> = targets.iter().collect();
+    let names: HashSet<&str> = targets.iter().map(|q| q.name()).collect();
+    let mut out: Vec<CallEdge> = Vec::new();
+    for edges in graph.forward.values() {
+        for e in edges {
+            let CallTarget::Bare(name) = &e.target else { continue };
+            let hit = if e.candidates.is_empty() {
```
vlsi, line 251: The candidate test here ignores `e.receiver`, which is populated on the edge and is the one discriminating signal available without type inference. Since the candidate set is keyed on the bare name, every unresolved `close()` in a project carries `PgConnection::close` among its candidates and lands in this list.

Measured on pgjdbc (~300k lines of Java):

| target | resolved | unattributed |
| --- | --- | --- |
| `CodecContextBuilder.prefersJavaTime` | 0 | 1 |
| `CodecContextBuilder.registry` | 2 | 5 |
| `TypeDescriptor.getOid` | 4 | 112 |
| `PgConnection.close` | 0 | **1124** |

The first three rows of that last section, with the source line pulled by `sed`:

```
benchmarks/.../LargeObjectRead.java:87     ->  os.close();          // OutputStream
benchmarks/.../LargeObjectRead.java:88     ->  lo.close();          // LargeObject
benchmarks/.../FinalizeConnection.java:76  ->  connection.close();
```

Same shape in this repository: `sb callers 'CliError.new'` reports 1 resolved caller and 1085 unattributed, whose candidate sets hold the nine project symbols named `new`. The sites are `Vec::new()`, `String::new()`, and friends. MCP is affected identically — `tools/call callers` returns `unattributed_total: 1085` with 199 rows in the body, straight into an agent's context.

Capping the list does not fix this on its own. The header prints `1124 unresolved call site(s) naming 'PgConnection.close' (... possible additional callers)`, and that number is short enough to be quoted as a rename cost — the opposite of what #31 was for.

Four fixes, cheapest first:

1. **Show the receiver.** It is already on `CallEdge` and appears in neither the text renderer nor `JsonUnattributed`. Free, and it turns an unreadable list into one a reader can triage at a glance.
2. **Filter the part that needs no inference.** When the receiver resolves to a project type, require it to match the target's enclosing type. Pass A in `resolve.rs` already applies exactly this rule; it is just not reused here.
3. **Gate the section on how discriminating the name is.** `symbol_table["close"].len()` is one lookup. When it comes back large, no edge naming that symbol is evidence about any single one of them — print one line saying so rather than a list. Consider phrasing the header with the reason: `receiver type unknown; 'close' is declared by 39 symbols in this project`.
4. **The real fix, for a follow-up: record declared types of locals and parameters.** In Java, C#, TypeScript, Go, and Rust the type sits next to the declaration — `ResultSet rs = ...`, `try (LargeObject lo = ...)`, `void f(PgConnection c)`. That is one more field in the walk the adapters already perform, not type inference. On the 400 sites I sampled there were 68 distinct receivers, led by `rs` (96), `stmt` (56), `st` (34), and `pstmt` (29); only 39 of 395 had a connection-shaped name, so this would cut roughly 90% even before the type is checked.

src/calls/cli.rs
```diff
@@ -110,16 +94,93 @@ pub fn run_callers(
...
+            "# note: {} unresolved call site(s) naming '{}' hidden (--hide-ambiguous); they may be additional callers",
+            unattributed_hidden,
+            target
+        );
+        unattributed.clear();
+    }
+    // The unattributed section gets whatever display budget the resolved
+    // hits and type groups left; its header carries its own true total, so
+    // nothing is silently cut.
+    let unattributed_total = unattributed.len();
```
vlsi: Two smaller problems with how the section is bounded.

**It shares `--limit` with the resolved hits**, so the emptier the resolved list, the more noise arrives. `callers PgConnection.close --limit 200` prints 0 resolved callers and 200 unattributed ones. The dependency runs the wrong way: a target the resolver handled badly should not get *more* display budget. The section wants its own, smaller cap.

**Nothing orders the section by relevance.** `unattributed_callers` sorts by `(file, line, source)`, so evidence strength plays no part. Sorting by `candidates.len()` ascending would put the useful rows first — measured on pgjdbc, `CodecContextBuilder.registry` edges carry 2 candidates each, `TypeDescriptor.getOid` mostly 3, and `PgConnection.close` between 7 and 39. Rendering the section last is already right, but position only helps a reader who stops early; it does nothing for a JSON consumer reading `unattributed_total`.

src/lib.rs
```diff
@@ -956,9 +1249,34 @@ pub fn run() {
             json,
             compact,
         } => {
-            let results = walk_and_parse(paths, None);
+            let paths = require_paths("implements", paths, *json);
```
vlsi, line 1333: `require_paths` makes `PATHS` mandatory here, but clap still declares it optional, so the help and the runtime disagree:

```
$ sb implements --help
Usage: sb implements [OPTIONS] <TARGET> [PATHS]...

$ sb implements CodecContextBuilder
error: `implements` received no paths (0 arguments after the subcommand)
exit=2
```

Rejecting is the right call. On 3.0.0 that same command exits 0 with `# 0 match(es)` while the type does have an implementor (`PgCodecContext.OfflineBuilder`), which is the #33 failure mode exactly. Two things to fix, though.

The message describes a different problem than the one that occurred. `0 arguments after the subcommand`, followed by advice about a `$(...)` substitution, sends the reader looking for a shell bug — but one argument was passed, and only the path list, which reads as optional, was left out.

Marking `paths` as required in clap fixes both at once: `--help` would say so, and clap's own message would name the missing argument instead of guessing at a cause. The alternative is the fallback this PR already gives the MCP `run` tool — default to `.`.

src/calls/resolve.rs
```diff
@@ -89,46 +89,186 @@ pub fn run_with_table(
...
+                            }
+                        }
+                        if let Some(rel) = rel {
+                            let want_tail = format!("::{}::{}", rest, raw.bare_name);
+                            let caller = raw.source.0.as_str();
+                            return match rel {
+                                // `crate::P` anchors at the file segment
+                                // *only* — an intermediate scope declaring
+                                // the same path must not shadow the root.
+                                SelfRel::Crate => {
```
vlsi, line 212: `crate::` anchors at the crate root, which is `src/lib.rs` or `src/main.rs`, not at the caller's file. Taking the caller's first qn segment makes the two agree only when the caller happens to live in the crate root.

Repro, a three-file `cargo` layout:

```rust
// src/inner.rs
pub struct Foo;
impl Foo { pub fn method() {} }                                    // the real target

// src/other.rs
pub mod inner { pub struct Foo; impl Foo { pub fn method() {} } }   // decoy
pub fn caller() { crate::inner::Foo::method(); }
```

```
$ sb callees 'other.rs:caller' .
src/other.rs:11    src/other.rs::inner::Foo::method  (Exact)     <- the decoy
$ sb callers 'src/inner.rs:method' .
# 0 caller(s)
```

`main` mis-binds this too, through `local_qn_by_name`, so it is not a regression. But note what the arm can do: it fires only when the caller's own file declares a path matching the crate-relative one, which is precisely the case where the answer is wrong. With no decoy present, the `find` misses and the edge falls through to pass B/C correctly. So the arm as written is net-negative, and the guarantee in 6c38348 — "an intermediate scope declaring the same path cannot shadow the root" — does not hold across files. The regression test covers only the same-file case. Gating the arm on the caller's file being `lib.rs` or `main.rs` would make it true.

`self::` and `super::` behaved correctly in my checks, including the ancestor-decoy case.

README.md
```diff
@@ -441,6 +441,21 @@ changes, so downstream tooling can guard on it:
 | `ast-bro.index-stats.v1` | `index --stats --json` |
 | `ast-bro.run.v1` | `run --json` |
 | `ast-bro.squeeze.v1` | `squeeze --json` |
+| `ast-bro.error.v1` | any rejected call under `--json` (on stderr) |
+
+---
+
+## CLI contract
+
+One rule set for every subcommand, so a consumer never needs a per-command table:
```
vlsi, line 450: `run`, `find-related`, and `search` do not follow the contract this section describes, and "one rule set for every subcommand" is the sentence an agent will act on:

```
$ sb run -p 'fn $A(){$$$}' --json /nope
{"schema":"ast-bro.run.v1","matches":[],"error_count":0}      # on stdout
exit=1

$ sb find-related /nope/f.rs:3 --json
{"schema":"ast-bro.related.v1", ... "results":[]}
exit=0

$ sb search 'zzz' ./nosuchdir --json
{"schema":"ast-bro.search.v1", ... "results":[]}
exit=0
```

`run` is the worst of the three: a rejected call gets exit 1 rather than 2, plus a valid-looking empty payload on stdout. Both halves contradict the paragraph above — exit 1 means internal failure, and a rejection is supposed to leave stdout empty. The MCP side of `run` did get `resolve_paths_for_mcp` in this PR; the CLI side did not.

Either extend `require_paths` to these three, or drop "every subcommand" and name the exceptions here. A contract that silently holds for most commands is harder to rely on than one that is explicitly partial.

vlsi (commented):
Reviewed at 72a53b5. I built the branch and ran the binary against two real repositories: this one, and pgjdbc (~300k lines of Java) where I had an existing API-naming task to drive it end to end. `cargo test --release --no-fail-fast` passes except `deps_e2e::go_module_prefix_strips_correctly`, which is red on `main` too.

The error contract holds up on real code. I re-measured every case in the diagnostics table I keep for this tool — `map /nope.java`, `surface /nope`, `show <file> NoSuchSymbol`, `implements NoSuchSymbol <path>`, `callers NoSuchSymbol`, `deps /nope.java`, `map` with no paths, and an unknown flag — and all eight now behave identically: exit 2, empty stdout, message on stderr. The table collapsed into one sentence, which is the point of #33/#36. The `map`/`digest` merge also came through clean: `map --preset digest --json` is byte-identical to `digest --json`, and `--max-members` text/JSON parity matched exactly on four files I checked.

Two things I would fix before merge, plus three smaller ones.

**Blocking — the unattributed section is unusable on common method names.** Details inline on `src/calls/traverse.rs`. On pgjdbc, `callers PgConnection.close` returns 0 resolved callers and 1124 unattributed; the first three are `connection.close()`, `os.close()` on an `OutputStream`, and `lo.close()` on a `LargeObject`. In this repository, `callers CliError.new` returns 1 resolved and 1085 unattributed — every `Vec::new()` and `String::new()` in the tree.

**`implements <TARGET>` now rejects a call that `--help` says is valid.** Details inline on `src/lib.rs`.

For the record, on precise targets the section does exactly what #31 promised, and that is worth keeping. On pgjdbc, `callers CodecContextBuilder.registry` reports 2 resolved (Inferred) plus 5 unattributed against 8 hits for `git grep '\.registry('`, the eighth being the declaration. `CodecContextBuilder.prefersJavaTime` reports 0 plus 1, and grep finds exactly one real call site. Fluent chains through an interface-typed receiver still do not resolve, but they no longer vanish silently. The fix is to gate the section on whether the name discriminates, not to remove it.

Three smaller items I did not anchor inline:

- `digest --json` drops `docs` from `ast-bro.map.v1` without a schema bump. Measured on `src/core.rs`: 18 `docs` keys and 55,586 bytes on `main`, 0 keys and 16,675 bytes here. The README documents the behavior, but the caller never opted in — the preset did — and the schema constant is unchanged, so a consumer indexing `docs` breaks with no signal to guard on. A release-note line would cover it; a `v2` would be safer.
- `callees <Type>` did not get the `--limit` fix that `callers <Type>` got in 45312b0. Ancestor groups are neither bounded by `--limit` nor counted in `total`, so `truncated` in `ast-bro.callees.v1` describes only `matches`. Inheritance chains are short, so this is a consistency gap rather than a volume problem.
- `callers`, `impact`, and `reverse-deps` now walk with `usize::MAX` to compute a true total. At the default `--depth 1` that costs nothing; at `--depth 5` on a large repository it is the entire reverse cone, where the old code stopped at 200 hits. The trade-off looks right, but `--limit` no longer bounds work, and the docs still read as though it does.

src/calls/cli.rs
```diff
@@ -201,6 +269,25 @@ pub fn run_callees(
...
+    }
+    // Same rationale as run_callers: the frontier note only accompanies an
+    // explicitly deepened walk; JSON always carries the flag.
+    if frontier_truncated && depth > 1 {
+        eprintln!(
+            "# note: --depth {} reached with unexplored edges beyond it; raise --depth to walk further",
+            depth
+        );
+    }
+    if total > all_edges.len() {
```
vlsi: Follow-up from re-measuring the `--limit` reporting across all four commands: `callees` is the only one whose **text header still prints the post-cap count**, so the #32 fix does not reach the surface most readers actually look at.

```
$ sb callees 'PgCodecContext.OfflineBuilder.build' --limit 2
# 2 callee(s) for '...::OfflineBuilder::build':          <- reads as complete
```
```
stderr: # note: 7 callee(s) total; showing 2 (raise --limit to see the rest)
json:   {"total": 7, "truncated": true, "matches": [ ...2 items... ]}
```

`callers` under the same conditions prints `# 10 caller(s) for 'x' (showing 3; raise --limit to see the rest)`. The asymmetry contradicts two claims this PR adds to its own docs — `wiki/architecture.md`: "`callers`/`callees`/`impact`/`reverse-deps` headers report the true total when `--limit` trims the display", and `SKILL.md`: "truncation is always reported, never silent". Neither holds for `callees` on stdout, and stdout is where `callees` gets read. Piping it to a file or reading it in a terminal that drops stderr leaves no trace that anything was cut.

The machinery is already in place. `render_callees_text` (`src/calls/render.rs:293`) calls `header_line("callee", kept.len(), ...)`, and `header_line` is now a thin wrapper over `header_line_suffixed`; passing the true total plus `section_suffix(total, kept.len())` gives exactly what `render_callers_text` already produces. `render_callees_text_extended` has the same gap at `src/calls/render.rs:202`, where the header is built from `total_ancestors + kept_flat.len()`.

One thing to settle while you are in there: `kept.len()` is also filtered by `include_external`, whereas the JSON `total` is `all_edges.len()` from before that filter. Under `--hide-external` the text header, the stderr note, and `total` therefore count three slightly different populations.

vlsi (commented):
One more, found while re-measuring the cap reporting on real repositories.

aeroxy:
@coderabbitai full review

aeroxy:
@coderabbitai full review

aeroxy:
@coderabbitai full review

aeroxy:
@coderabbitai full review

aeroxy:
@coderabbitai full review

PR #44, opened 2026-08-04
fix: multi-file/dir/glob targets, frontmatter, path repair, coverage

aeroxy:
Make `show` accept a file, directory, quoted glob, or explicit file list. Partition targets vs symbols by what exists on disk so an unquoted shell glob is recovered instead of silently searching only the first file. Multi-file answers carry a coverage header and a display --limit (default 20); single-file output stays uncapped and headerless. Schema bumps to ast-bro.show.v2 (files array). Shared engine used by CLI and MCP.

Detect leading YAML frontmatter in Markdown (BOM-tolerant, trailing whitespace on fences, `...` closer) as a Frontmatter declaration so frontmatter-only cards no longer map empty; `show … frontmatter` and digest include it.

Path-not-found errors offer verifiable repairs only: rejoin unquoted paths with spaces, or the same basename elsewhere under the normal filter pipeline (capped, ignore-aware). Wire into require_paths, show, squeeze, find-related, and MCP. Misread paths in the symbol slot are diagnosed as paths, not absent symbols.

`run` honest zeros report files_scanned on stderr and in JSON so a narrow --glob cannot be mistaken for a real empty result.

## Summary by CodeRabbit

- **New Features**
  - `show` now supports multiple files, directories, and quoted globs.
  - Added result limits, scan coverage, match totals, and enhanced text/JSON output.
  - Markdown YAML frontmatter is included in map, digest, and show results.
  - Added helpful suggestions for mistyped or split paths.
  - `run` output reports scanned-file counts and clearer zero-match diagnostics.

- **Bug Fixes**
  - Prevented duplicate results and repeated rewrites when targets overlap.

- **Documentation**
  - Updated CLI, MCP, and architecture documentation for expanded capabilities and the `show` JSON v2 format.


aeroxy:
@coderabbitai full review

aeroxy:
Triaged all ten. Three were real (8bbb203); seven weren't.

Fixed
run's empty-result line overstated coverage — the one real bug. attempted_files counts files picked up including ones that then errored (oversize, unreadable, uncompilable pattern). Reproduced with a 5 MB file:


before:  no matches for "X" (2 file(s) scanned)          ← only 1 was actually read
after:   no matches for "X" (2 file(s) scanned, 1 errored, so this zero is not exhaustive)
Same overstated-coverage failure that line was added to prevent, one level down. A clean zero is unchanged so nothing gets hedged unnecessarily; JSON already separated files_scanned from error_count. Two tests added.

MCP show.limit had no minimum — 19 integers in the schema have one; this was one of only two without, so a negative failed at deserialization instead of validation. minimum: 0, preserving limit: 0.

starts_with('/') in my new test → Path::is_absolute(). Plainer way to say the same thing.

Skipped
Markdown frontmatter overlap — not reproducible. The premise is sound (title: x / --- is a setext-h2 candidate) but tree-sitter-md doesn't emit one. Tight, frontmatter-only, and BOM-prefixed documents all produce exactly one frontmatter decl and no overlapping heading.

find-related repair on a directory — the hint is correct. A directory named foo.rs gives "not a file: foo.rs" plus "found the same file name at: sub/foo.rs" — a verified on-disk repair, which is exactly what path_repair exists for. Suppressing it would remove a good hint. Exit stays 2.

MCP show target list — real asymmetry, deliberate. The CLI takes a target list; MCP takes one path. The code says why ("Kept as one field so the tool schema is unchanged"). Worth closing, but it's a schema change, not a minimal fix. Say the word.

README fence — 29 of the README's fences are bare. Tagging one breaks the convention rather than restoring it.

Windows separators — no CI, no cfg(windows), no Windows distribution anywhere in the repo. (I took is_absolute() because it's clearer code, not for portability.)

BOM fixture and swapping the Cargo.toml fixture — coverage suggestions for behavior that's already correct. I verified BOM handling works; cargo test guarantees the cwd for the other.

636 tests pass, clippy clean.

aeroxy:
@coderabbit full review

aeroxy:
All 352+ tests pass, clippy clean.

**Fixed**

*   [src/mcp/tools.rs:1385](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/mcp/tools.rs#L1385) — the MCP search zero-message quoted `files_scanned` alone while that count includes errored files, so a partial scan read as complete. Now mirrors the CLI wording (`N file(s) scanned, M errored, so this zero is not exhaustive`). The rewrite-mode zero already guarded on `error_count == 0`, so it needed nothing.
*   [src/mcp/tools.rs:1103](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/mcp/tools.rs#L1103) — comment now states what `files_scanned` counts (reached files with a supported language, errored ones included). No schema `description` exists for the field, so there was nowhere else to apply the wording.
*   [src/mcp/tools.rs:690](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/mcp/tools.rs#L690) — the not-a-target rejection now appends `path_repair::hints` when the path is missing, matching the CLI; an existing-but-unparseable file keeps the plain message.
*   [src/mcp/tools.rs:1313](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/mcp/tools.rs#L1313) — MCP `RewriteDoc` was the only run document without `schema`. Added.
*   [tests/show\_targets.rs](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/tests/show_targets.rs) — two inline `Command` blocks replaced with the existing `run_in`; `unparseable_single_target_still_says_unsupported` now builds its `Cargo.toml` path from `CARGO_MANIFEST_DIR`.
*   [tests/overlapping\_targets.rs:143](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/tests/overlapping_targets.rs#L143) — nested-target test now asserts `files_scanned == 2`.
*   [tests/markdown\_frontmatter.rs](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/tests/markdown_frontmatter.rs) — added a `bom.md` fixture and `a_bom_does_not_push_the_fence_off_line_one`, covering the BOM branch at [src/adapters/markdown.rs:70](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/adapters/markdown.rs#L70). The coverage assertion now derives its count from the id list instead of hardcoding it.

**Skipped**

*   **show\_targets.rs:87-90 "quote it" assertion** — the claim is wrong. That string comes from the intended expanded-glob note at [src/lib.rs:1364](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/lib.rs#L1364) (`If the shell expanded an unquoted glob, quote it…`), which fires for exactly this all-files argument list. It is not `path_repair`'s hint, and the test exercises what it says it does.
*   **Removing MCP `capped`/`cap_limit`/`errors` to match the CLI docs** — those fields carry real information the CLI has no need for (MCP enforces match/file caps and has no stderr for the error list). Dropping them loses data; I aligned the one field that was a genuine gap (`schema`).
*   **`partition` caching `is_target`** — duplicate stat calls over a handful of args, short-circuited in the common case. No behavior impact.
*   **`render_json` typed struct** — serializing an already-built `serde_json::Value` cannot fail, so the "silent fallback" is unreachable. Pure style churn on a v2 schema.
*   **Centralizing `bin`/`run`/`run_in` into `tests/common/mod.rs`** — three ~15-line helpers across three files; a shared module adds indirection for no correctness gain.

aeroxy:
Fixed at [src/mcp/tools.rs:690-707](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/mcp/tools.rs#L690-L707). The branch now mirrors [src/lib.rs:1298-1312](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/lib.rs#L1298-L1312) verbatim in wording:

*   exists but no adapter → `unsupported file type for \`show\`: {path}\`, no hint (nothing on disk to repair)
*   absent → `path not found: {path}` plus `path_repair::hints`

I dropped the glob nuance rather than keeping it: a glob that matched nothing never `exists()`, and the CLI reports it as `path not found` too (asserted by `a_glob_matching_nothing_is_a_path_error_not_a_symbol_error`). MCP-only wording there would reintroduce the drift this fix removes.

Full suite passes, clippy clean.

aeroxy:
Valid — and it affected the CLI too, not just MCP: [src/lib.rs:1310](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/lib.rs#L1310) calls `hints(&raw)` on the same unmatched-glob path. So I took Option B.

**Fix** — [src/path\_repair.rs:119](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/path_repair.rs#L119): `elsewhere` returns `None` when the basename carries `*`, `?`, or `[`, alongside the existing extensionless-bare-word guard. Every caller (CLI `show`, `require_paths`, `require_path`, `find-related`, MCP) now skips the walk, and the reasoning is stated once where the walk is.

`rejoined` is untouched, so the one glob repair that *is* verifiable still fires: `map The Sorting Bureau/*.py` split on spaces, where [expand\_pattern](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/path_repair.rs#L108) proves the rejoin. That's the case the guard had to avoid breaking.

**Tests** — [src/path\_repair.rs](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/path_repair.rs), two added:

*   `does_not_scan_for_a_glob_basename` — `src/*.nope`, `*.rs`, `src/widget[1].rs` all return `None` from `elsewhere`, and `hints` produces nothing for a lone glob.
*   `a_split_glob_still_gets_its_rejoin` — the spaces-plus-glob case still reports `rejoined … Bureau/*.py`, which is what would regress if the guard were placed in `hints` instead.

A metachar in the directory part only (`src/*/widget.rs`) still gets its repair — the basename is literal there, so finding `lib/widget.rs` remains a real suggestion.

Full suite passes, clippy clean.

aeroxy:
@coderabbitai full review

aeroxy:
**Fixed**

*   [README.md:483-488](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/README.md#L483-L488) (qodo) — correct. I had lumped `unmatched` in with the counters and attributed it to the text header. It's an array of symbols that matched nothing, and on the CLI it surfaces as a stderr note, not in the header. Reworded to split the two and say where `unmatched` actually appears.
    
*   [src/path\_repair.rs:52-62](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/path_repair.rs#L52-L62) (coderabbit) — real bug, and the code contradicted its own doc comment (`MAX_REPAIRED_ARGS` is documented as a cap on *missing* arguments). `.take(3)` ran before the `exists()` skip, so existing args spent the budget. Reachable through [src/lib.rs:1310](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/lib.rs#L1310), which passes every positional: `show Cargo.toml a.md b.md src/typo.rs Sym` produced no hint for `src/typo.rs` because the three existing files consumed the cap. Filter now precedes the cap. Locked in by `existing_args_do_not_spend_the_repair_budget`, which fails on the old order.
    

**Skipped**

*   **`_digest_markdown` frontmatter vs the depth guard** — not a bug, on two independent grounds. Frontmatter is pushed as a top-level declaration ([src/adapters/markdown.rs:20](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/adapters/markdown.rs#L20)), so it is only ever seen at `depth == 0`, and the guard is `depth > max_heading_depth` — `0 > 0` is false, so even at zero the frontmatter branch is reached. Separately, no code path sets `max_heading_depth` to 0: the default is 3 and all three call sites pass 3, with no flag to change it. I confirmed the claimed failure case empirically — a frontmatter-only file renders its block under `digest`. Moving the branch above the guard would be dead code guarding an unreachable state.

Full suite passes, clippy clean.

aeroxy:
@coderabbitai full review

aeroxy:
**Fixed** — finding 1. Added `path_glob::pattern_matches_any` ([src/path\_glob.rs:17-26](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/path_glob.rs#L17-L26)), which short-circuits on the first match instead of collecting every one.

I applied it at both boolean call sites, not just the one you named: [src/path\_repair.rs:127](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/path_repair.rs#L127) (`rejoined`) and [src/show.rs:118](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/show.rs#L118) (`is_target`). The second matters more — `is_target` runs on the normal path, twice per argument inside `partition`, so `show '**/*' Sym` was building the full expansion twice just to ask whether it matched anything. The three call sites that genuinely need the list (manifest globs, workspace members/excludes) still use `expand_pattern`.

Verified both behaviors survive: the rejoined-glob hint still fires (`quote it: "The Sorting Bureau/*.py"`), and a quoted glob still resolves as a target.

**Skipped** — finding 2, with a measurement.

The worst case you describe costs **451 ms** total (`map zz_no_such_widget.rs` from `$HOME`, no marker present) against a **10 ms** baseline in a small directory. So the walk is real, but two things argue against both proposed guards:

1.  **It isn't new, and isn't specific to the relative-arg change.** A bare relative filename has an empty parent, so the climb fell through to `PathBuf::from(".")` — the cwd — before my change too. `map typo.rs` from `$HOME` walked the cwd already; the change extended that to args whose parent exists, which is the same anchor.
2.  **Both options undo a property just fixed.** Option 1 makes the hint silently absent in `$HOME`/`/`, and option 2 restores the nearest-existing-ancestor anchor for broad roots — that's precisely the narrow-search silence deepseek reported and I fixed last round. "No hint" reading as "no such file anywhere" is the failure mode the feature exists to prevent, so trading it for latency that's already bounded is the wrong direction.

If the 20k bound turns out to be too generous on network mounts, the honest lever is `MAX_ENTRIES` itself ([src/path\_repair.rs:30](vscode-webview://19il7713bho0faifevo5un4v4ic12flfgncsf3ualuj3t32gvrro/src/path_repair.rs#L30)) — one tuned constant with a stated rationale — rather than a cwd special case that has to guess which directories are "broad" and gets it wrong for anyone whose project *is* `$HOME`.

Full suite green, clippy clean.

PR #50, opened 2026-08-09
fix: deliver a map for a Read the host refuses outright

### LINKED ISSUE #34: hook: substitution reads as a tool failure, and the explanation lands last

vlsi:
## Why

The hook replaces a large `Read` with the file's structure map. A file large enough that Claude Code refuses to read it at all never gets that far: the host rejects the call on its own, and the agent is left holding `File content … exceeds maximum allowed tokens` and nothing else. That is the worst case to lose the map on — the bigger the file, the more the map is worth and the more a full read would have cost.

Those refusals also land on files the hook's own threshold never sees, because the two count different things: `--min-lines` counts lines, the host's limit counts bytes. A 90-line file of 355 KB trips the host and no threshold of ours.

Separately, the map itself was never bounded. A minified bundle is one line holding thousands of declarations, so 255 KB of it mapped to 207 KB — roughly twice the tokens the refusal was about, and on the refusal channel that same map replaced a read the agent never paid for.

## What

- **A map for a read the host refuses.** The installer registers the hook under `PostToolUseFailure` too, and a new response channel delivers the map through `hookSpecificOutput.additionalContext`. Verified on a live host: the event fires for a `Read` the host rejected over its size limit, and the context reaches the model.
- **On that event the line-count and range gates drop.** The read is not happening again, lines are not what refused it, and the host's own error tells the agent to retry with `offset`/`limit` — so gating that retry out would leave the hole open. On the refusal channel both gates stay, where a substitution would replace what the caller asked for.
- **The map is bounded, shedding detail before cutting anything.** Order follows what each level costs: doc comments and attributes first, which are detail about a declaration, then fields and private items, which are declarations. Measured on Apache Calcite's `SqlFunctions.java` (7771 hand-written lines): 142,692 bytes in full, 74,607 with all 920 declarations once docs and attributes go. Both markers name what is missing and the command that returns it, because a silently short map reads as a file that ends early.
- **`status` reports `hook partial`** when only some of the events an install writes are registered — the state every existing install is in after this change — with the remedy on stderr, where the repo's contract puts qualifications of a delivered answer.
- **No empty hook events left behind.** `json_hook::remove` drops an event key once it is empty, so installing or uninstalling no longer leaves `"PreToolUse": []` in a file the user owns.
- **Internals that make the above possible.** `decide` returns the bare map and `io` owns the framing, so one command serves three events with three notices; `Channel::is_additive()` is the single place that says whether a payload replaces something or is added to it.

A normal large `Read` behaves as before: `PreToolUse` refuses the call and carries the map — now bounded.

## How to verify

`cargo test` — full suite green, `clippy --all-targets` clean.

Gates: `failed_ranged_read_is_substituted_but_a_successful_one_is_not` asserts both directions of the range gate, `failed_read_substitutes_below_the_line_threshold` the line gate, `failed_read_still_needs_a_parseable_file` that only those two gates moved. Bounding: `a_large_hand_written_file_sheds_detail_instead_of_declarations` builds a fixture whose full map exceeds the cap and asserts nothing was truncated, the shed is named, and the first and last declarations survive; `a_minified_bundle_is_trimmed_and_points_at_the_whole_map` covers the backstop. Registration: `install_hook_registers_pre_tool_use_and_the_failure_event` also asserts `PostToolUse` is *not* registered. Status: `status_reports_partial_for_an_install_missing_the_failure_event` and `status_is_not_partial_after_install_completes_it`. Notices: `every_notice_keeps_its_command_on_one_line` catches a wrap inside the backticks that would comment out the symbol argument.

By hand, on the file the host refuses:

```console
$ echo '{"hook_event_name":"PostToolUseFailure","tool_name":"Read","tool_input":{"file_path":"huge.rs"}}' | ast-bro hook --protocol claude-code --min-lines 200
{"hookSpecificOutput":{"hookEventName":"PostToolUseFailure","additionalContext":"# ast-bro supplies this structure map, with line ranges, because the Read did…"}}
```

Installing over a settings file leaves the user's own hooks untouched; uninstalling leaves no empty keys.

## Note on #34

#34 asks for a channel that reports the substitution as a success rather than a failed tool call. That channel does not exist for the built-in `Read` on Claude Code 2.1.223: `hookSpecificOutput.updatedToolOutput` replaces the result of an MCP tool and is ignored for `Read` and `Bash`. Measured with a hook returning `updatedToolOutput` and `additionalContext` in one object — the second lands every time, the first never does — with workspace trust, field name, and envelope parsing ruled out ([evidence](https://github.com/aeroxy/ast-bro/issues/34#issuecomment-5233506130)).

So `PreToolUse` remains the registered channel, and `Channel::Replace` stays in the code unregistered, carrying that measurement and reachable from a hand-written `PostToolUse` entry for anyone whose host honors the field. This PR does not close #34.

Refs #34.

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,14 +331,16 @@ call, so the map arrives beside the error rather than in place of
 anything.
 
 Whatever the channel, the map is capped at 64 KB. A map that would run
-over sheds detail first, in order of what it costs: doc comments and
-attributes, then fields and private items, then members past 50 per
-type. Only when that still does not fit is the map trimmed line by line
-— a minified bundle, where the size is sheer declaration count, or a
-single generated declaration too wide to fit at all. A trim drops a
-declaration's members along with it rather than filing them under the
-previous one. Either way the payload ends with a note saying what is
-missing and the `ast-bro map` command that returns it.
+over can shed detail — doc comments and attributes first, then fields and
+private items, with a per-type member cap as a last backstop that rarely
+comes up — or keep the detail and drop whole entries instead. Which one you get is decided by how many
+declarations each delivers, not by which fits first: a mostly-private
+file would fit the moment private items went, and hand you the public
+few, so it is trimmed at a richer level instead. A trim drops a
+declaration together with its doc comment and its members, rather than
+leaving them to read as someone else's. Either way the payload ends with
+a note saying what is missing and the `ast-bro map` command that returns
+it.
 
 ### Claude Code subagent shadowing
 
--- a/README.md
+++ b/README.md
@@ -339,10 +339,11 @@ went, and hand you the public few, so it is trimmed at a richer level
 instead.
 
 A trim spends the 64 KB in the order that ladder ranks: whatever survives
-furthest down it goes in first, so a file's public surface arrives whole
-before its private helpers get the rest. Filling in file order instead
-handed back 1733 private items and none of that file's 500 public ones,
-because the publics ran to the end of a file twice the budget. What a
+furthest down it goes in first, so a file's public surface is what the
+budget buys before anything else — all of it, unless the public surface
+alone is over the cap. Filling in file order instead handed back 1733
+private items and none of that file's 500 public ones, because the
+publics ran to the end of a file twice the budget. What a
 trim drops, it drops entire — a declaration leaves with its doc comment
 and its members, rather than leaving them to read as someone else's. The
 payload then ends with a note saying what is missing and the
--- a/README.md
+++ b/README.md
@@ -324,23 +324,29 @@ the read the host refuses outright — measured against its file-size
 limit, 256 KB on Claude Code 2.1.223. That limit counts bytes, not lines,
 so it catches files no `--min-lines` threshold does: a 90-line file of
 355 KB trips it. Without this the agent gets a bare error and nothing
-else. That event cannot replace a result, only
-add context, which is all the map needs when the result carries no file
-contents. Nothing is blocked on that path: the host already failed the
-call, so the map arrives beside the error rather than in place of
-anything.
+else. That event cannot replace a result, only add context, which is all
+the map needs when the result carries no file contents. Nothing is
+blocked on that path: the host already failed the call, so the map
+arrives beside the error rather than in place of anything.
 
 Whatever the channel, the map is capped at 64 KB. A map that would run
 over can shed detail — doc comments and attributes first, then fields and
 private items, with a per-type member cap as a last backstop that rarely
-comes up — or keep the detail and drop whole entries instead. Which one you get is decided by how many
-declarations each delivers, not by which fits first: a mostly-private
-file would fit the moment private items went, and hand you the public
-few, so it is trimmed at a richer level instead. A trim drops a
-declaration together with its doc comment and its members, rather than
-leaving them to read as someone else's. Either way the payload ends with
-a note saying what is missing and the `ast-bro map` command that returns
-it.
+comes up — or keep the detail and drop whole entries instead. Which one
+you get is decided by how many declarations each delivers, not by which
+fits first: a mostly-private file would fit the moment private items
+went, and hand you the public few, so it is trimmed at a richer level
+instead.
+
+A trim spends the 64 KB in the order that ladder ranks: whatever survives
+furthest down it goes in first, so a file's public surface arrives whole
+before its private helpers get the rest. Filling in file order instead
+handed back 1733 private items and none of that file's 500 public ones,
+because the publics ran to the end of a file twice the budget. What a
+trim drops, it drops entire — a declaration leaves with its doc comment
+and its members, rather than leaving them to read as someone else's. The
+payload then ends with a note saying what is missing and the
+`ast-bro map` command that returns it.
 
 ### Claude Code subagent shadowing
 
--- a/README.md
+++ b/README.md
@@ -300,9 +300,45 @@ ast-bro status
 
 Supported targets: `claude-code`, `gemini`, `tabnine`, `cursor`,
 `aider`, `codex`, `copilot`, `opencode`. Claude Code, Gemini, and Tabnine also get
-a tool-call hook that intercepts `Read` on supported source files when
-they exceed `--min-lines` (default 200) and substitutes the map output.
+a tool-call hook that intercepts `Read` on supported source files above
+`--min-lines` (default 200) and substitutes the map output. On Claude
+Code it also covers a read the host refuses outright, at any line count.
 The other targets do not install a read-interceptor hook.
+
+The substitution reaches the agent as a blocked tool call whose first
+line says that nothing failed. For a read that has not run yet, that is
+the only channel available: the map has to arrive instead of the file, so
+the read must be refused, and a refusal is what the host reports.
+
+Claude Code has a second event that would report a success instead,
+`PostToolUse` with `hookSpecificOutput.updatedToolOutput`, and
+`ast-bro install` does not register it. Measured on Claude Code 2.1.223:
+that field replaces the result of an MCP tool and is ignored for the
+built-in `Read`, so registering it would deliver no map and send the
+whole file to the model. Registering a `PostToolUse` entry by hand does
+nothing useful for `Read` on that version, for the same reason. See
+[issue #34](https://github.com/aeroxy/ast-bro/issues/34).
+
+On Claude Code the hook also registers under `PostToolUseFailure`, for
+the read the host refuses outright — measured against its file-size
+limit, 256 KB on Claude Code 2.1.223. That limit counts bytes, not lines,
+so it catches files no `--min-lines` threshold does: a 90-line file of
+355 KB trips it. Without this the agent gets a bare error and nothing
+else. That event cannot replace a result, only
+add context, which is all the map needs when the result carries no file
+contents. Nothing is blocked on that path: the host already failed the
+call, so the map arrives beside the error rather than in place of
+anything.
+
+Whatever the channel, the map is capped at 64 KB. A map that would run
+over sheds detail first, in order of what it costs: doc comments and
+attributes, then fields and private items, then members past 50 per
+type. Only when that still does not fit is the map trimmed line by line
+— a minified bundle, where the size is sheer declaration count, or a
+single generated declaration too wide to fit at all. A trim drops a
+declaration's members along with it rather than filing them under the
+previous one. Either way the payload ends with a note saying what is
+missing and the `ast-bro map` command that returns it.
 
 ### Claude Code subagent shadowing
 
@@ -313,7 +349,7 @@ containing the full ast-bro prompt.
 
 When you run `ast-bro install --target claude-code`, you get:
 - `CLAUDE.md` — main agent prompt (global or local per-repo)
-- `.claude/settings.json` — `Read` tool hook
+- `.claude/settings.json` — `Read` hooks on `PreToolUse` and `PostToolUseFailure`
 - `.claude/agents/Explore.md` — Explore subagent with the prompt injected
 
 This solves the "why doesn't my subagent use ast-bro?" problem — subagents
--- a/src/core.rs
+++ b/src/core.rs
@@ -620,17 +620,31 @@ fn _size_label(line_count: usize) -> &'static str {
 
 
 pub fn render_map(result: &ParseResult, opts: &MapOptions) -> String {
-    let mut lines = vec![_format_file_header(
+    render_map_units(result, opts).join("\n")
+}
+
+/// The map as the units it is assembled from, rather than as one string.
+///
+/// A unit is the file header, a warning, one declaration together with its doc
+/// comment, a `... +N more member(s)` marker, or a blank separator. Text is what
+/// [`render_map`] joins these into; a caller that has to *shorten* a map needs
+/// the units, because a physical line does not tell you what nests under what:
+/// a doc comment keeps the indentation it had in the source, so its continuation
+/// lines land at arbitrary columns, and a separator has no indentation at all.
+/// The unit's own indentation is the render prefix, and children are separate
+/// units indented one level deeper.
+pub fn render_map_units(result: &ParseResult, opts: &MapOptions) -> Vec<String> {
+    let mut units = vec![_format_file_header(
         &format!("# {}", result.path.display()),
         result,
     )];
     if let Some(warn) = _format_error_warning(result) {
-        lines.push(warn);
+        units.push(warn);
     }
     for decl in &result.declarations {
-        _render_decl(decl, opts, 0, &mut lines);
+        _render_decl(decl, opts, 0, &mut units);
     }
-    lines.join("\n")
+    units
 }
 fn _format_file_header(prefix: &str, result: &ParseResult) -> String {
     let counts = _collect_counts(&result.declarations);
@@ -739,8 +753,12 @@ fn _render_decl(decl: &Declaration, opts: &MapOptions, indent: usize, out: &mut 
 
     let prefix = "    ".repeat(indent);
 
+    // The declaration and its doc comment go out as one unit, so a consumer that
+    // drops the declaration cannot leave the comment behind to read as
+    // documentation of whatever follows it.
+    let mut unit: Vec<String> = Vec::new();
     if opts.include_docs && !decl.docs.is_empty() && !decl.docs_inside {
-        push_docs(&prefix, out);
+        push_docs(&prefix, &mut unit);
     }
 
     let attrs_prefix = if opts.include_attributes && !decl.attrs.is_empty() {
@@ -756,13 +774,13 @@ fn _render_decl(decl: &Declaration, opts: &MapOptions, indent: usize, out: &mut 
     };
 
     if decl.kind == Namespace {
-        out.push(format!(
+        unit.push(format!(
             "{}namespace {}",
             prefix,
             decl.name.magenta().bold()
         ));
     } else {
-        out.push(format!(
+        unit.push(format!(
             "{}{}{}{}",
             prefix, attrs_prefix, decl.signature, suffix
         ));
@@ -770,8 +788,9 @@ fn _render_decl(decl: &Declaration, opts: &MapOptions, indent: usize, out: &mut 
 
     if opts.include_docs && !decl.docs.is_empty() && decl.docs_inside {
         let inner_prefix = "    ".repeat(indent + 1);
-        push_docs(&inner_prefix, out);
+        push_docs(&inner_prefix, &mut unit);
     }
+    out.push(unit.join("\n"));
 
     // `--max-members` caps how many members a type renders; the remainder
     // is reported, never silently dropped (issues #37 / #32).
--- a/src/core.rs
+++ b/src/core.rs
@@ -620,29 +620,82 @@ fn _size_label(line_count: usize) -> &'static str {
 
 
 pub fn render_map(result: &ParseResult, opts: &MapOptions) -> String {
-    render_map_units(result, opts).join("\n")
+    render_map_units(result, opts)
+        .iter()
+        .map(|u| u.text.as_str())
+        .collect::<Vec<_>>()
+        .join("\n")
+}
+
+/// What a map is assembled from: one piece of text plus the structure that text
+/// does not carry.
+///
+/// [`render_map`] joins the text and drops the rest, which is all a reader needs.
+/// A caller that has to *shorten* a map needs the rest: which declaration a piece
+/// renders, and which declaration encloses it. Rendered text answers neither
+/// question. Indentation looks like an answer and is not one — a doc comment
+/// keeps the indentation it had in the source, so its continuation lines land at
+/// arbitrary columns, and a separator has no indentation at all.
+#[derive(Debug, Clone)]
+pub struct MapUnit {
+    /// The text [`render_map`] writes for this piece.
+    pub text: String,
+    /// What this piece is, and for a declaration, which one.
+    pub kind: MapUnitKind,
+    /// The declaration this piece nests under, `None` at the top level.
+    pub parent: Option<usize>,
+}
+
+/// What a [`MapUnit`] is.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MapUnitKind {
+    /// The file header or a parse warning: not a declaration, and a map missing
+    /// one reads as a map of a different file.
+    Preamble,
+    /// One declaration together with its doc comment.
+    ///
+    /// `id` numbers the declaration in pre-order over the whole tree, counting
+    /// the declarations [`MapOptions`] filtered out as well, so the number means
+    /// the same thing in every render of one file and two renders can be matched
+    /// against each other.
+    Declaration { id: usize },
+    /// A `... +N more member(s)` marker: a statement about the enclosing type
+    /// rather than a declaration of its own.
+    MoreMembers,
+    /// The blank line between items, which carries no structure.
+    Separator,
+}
+
+impl MapUnit {
+    fn preamble(text: String) -> Self {
+        Self {
+            text,
+            kind: MapUnitKind::Preamble,
+            parent: None,
+        }
+    }
+
+    fn separator() -> Self {
+        Self {
+            text: String::new(),
+            kind: MapUnitKind::Separator,
+            parent: None,
+        }
+    }
 }
 
 /// The map as the units it is assembled from, rather than as one string.
-///
-/// A unit is the file header, a warning, one declaration together with its doc
-/// comment, a `... +N more member(s)` marker, or a blank separator. Text is what
-/// [`render_map`] joins these into; a caller that has to *shorten* a map needs
-/// the units, because a physical line does not tell you what nests under what:
-/// a doc comment keeps the indentation it had in the source, so its continuation
-/// lines land at arbitrary columns, and a separator has no indentation at all.
-/// The unit's own indentation is the render prefix, and children are separate
-/// units indented one level deeper.
-pub fn render_map_units(result: &ParseResult, opts: &MapOptions) -> Vec<String> {
-    let mut units = vec![_format_file_header(
+pub fn render_map_units(result: &ParseResult, opts: &MapOptions) -> Vec<MapUnit> {
+    let mut units = vec![MapUnit::preamble(_format_file_header(
         &format!("# {}", result.path.display()),
         result,
-    )];
+    ))];
     if let Some(warn) = _format_error_warning(result) {
-        units.push(warn);
+        units.push(MapUnit::preamble(warn));
     }
+    let mut next_id = 0usize;
     for decl in &result.declarations {
-        _render_decl(decl, opts, 0, &mut units);
+        _render_decl(decl, opts, 0, None, &mut next_id, &mut units);
     }
     units
 }
@@ -734,7 +787,14 @@ fn _collect_counts(decls: &[Declaration]) -> std::collections::HashMap<&'static 
     out
 }
 
-fn _render_decl(decl: &Declaration, opts: &MapOptions, indent: usize, out: &mut Vec<String>) {
+fn _render_decl(
+    decl: &Declaration,
+    opts: &MapOptions,
+    indent: usize,
+    parent: Option<usize>,
+    next_id: &mut usize,
+    out: &mut Vec<MapUnit>,
+) {
     use DeclarationKind::*;
 
     let push_docs = |prefix: &str, lines_out: &mut Vec<String>| {
@@ -747,7 +807,14 @@ fn _render_decl(decl: &Declaration, opts: &MapOptions, indent: usize, out: &mut 
         }
     };
 
+    let id = *next_id;
+    *next_id += 1;
+
     if !_map_eligible(decl, opts) {
+        // Number the subtree anyway: a declaration's number has to name the same
+        // declaration whatever the options dropped, or two renders of one file
+        // cannot be matched against each other.
+        *next_id += _decl_count(decl);
         return;
     }
 
@@ -790,7 +857,11 @@ fn _render_decl(decl: &Declaration, opts: &MapOptions, indent: usize, out: &mut 
         let inner_prefix = "    ".repeat(indent + 1);
         push_docs(&inner_prefix, &mut unit);
     }
-    out.push(unit.join("\n"));
+    out.push(MapUnit {
+        text: unit.join("\n"),
+        kind: MapUnitKind::Declaration { id },
+        parent,
+    });
 
     // `--max-members` caps how many members a type renders; the remainder
     // is reported, never silently dropped (issues #37 / #32).
@@ -806,21 +877,24 @@ fn _render_decl(decl: &Declaration, opts: &MapOptions, indent: usize, out: &mut 
         if visible(child) && _counts_toward_member_cap(&child.kind) {
             if shown >= cap {
                 hidden += 1;
+                *next_id += 1 + _decl_count(child);
                 continue;
             }
             shown += 1;
         }
-        _render_decl(child, opts, indent + 1, out);
+        _render_decl(child, opts, indent + 1, Some(id), next_id, out);
     }
     if hidden > 0 {
-        out.push(
-            format!(
+        out.push(MapUnit {
+            text: format!(
                 "{}    ... +{} more member(s) (raise --max-members)",
                 prefix, hidden
             )
             .dimmed()
             .to_string(),
-        );
+            kind: MapUnitKind::MoreMembers,
+            parent: Some(id),
+        });
     }
 
     if indent == 0
@@ -829,8 +903,13 @@ fn _render_decl(decl: &Declaration, opts: &MapOptions, indent: usize, out: &mut 
             Class | Struct | Interface | Record | Enum | Namespace
         )
     {
-        out.push(String::new());
+        out.push(MapUnit::separator());
     }
+}
+
+/// Declarations under `decl`, `decl` itself excluded.
+fn _decl_count(decl: &Declaration) -> usize {
+    decl.children.iter().map(|c| 1 + _decl_count(c)).sum()
 }
 
 fn _clip_docs(docs: &[String], limit: usize) -> Vec<String> {
--- a/src/hook/claude_code.rs
+++ b/src/hook/claude_code.rs
@@ -1,22 +1,32 @@
-//! Claude Code PreToolUse hook protocol shim.
+//! Claude Code hook protocol shim, for both events that can carry a map.
 //!
-//! Claude Code's PreToolUse event JSON shape:
-//!   { "tool_name": "Read",
+//! Claude Code's tool-call event JSON shape:
+//!   { "hook_event_name": "PostToolUse",
+//!     "tool_name": "Read",
 //!     "tool_input": { "file_path": "...", "offset": ..., "limit": ... } }
 //!
-//! Response shape is shared with Gemini — see `super::io`.
+//! `hook_event_name` picks the response channel, so one command serves an
+//! entry registered under either event and `ast-bro install` can move the
+//! entry without changing the command string. Response shapes live in
+//! `super::io`.
 
 use std::path::PathBuf;
 
 use serde::Deserialize;
 
 use super::decide::DecideOpts;
-use super::event::ToolCallEvent;
+use super::event::{Channel, ToolCallEvent};
 use super::io::{dispatch, emit_pass_through, read_stdin};
 
 #[derive(Debug, Deserialize)]
 struct InputEvent {
     tool_name: String,
+    /// The event that fired, absent on a host too old to send it. Only
+    /// `PostToolUse` can replace a delivered result and only
+    /// `PostToolUseFailure` reaches a read the host rejected, so everything
+    /// else — including a missing value — falls back to refusing the call.
+    #[serde(default)]
+    hook_event_name: Option<String>,
     #[serde(default)]
     tool_input: ToolInput,
 }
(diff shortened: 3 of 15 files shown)


aeroxy:
> ## Why
> The hook replaces a large `Read` with the file's structure map. A file large enough that Claude Code refuses to read it at all never gets that far: the host rejects the call on its own, and the agent is left holding `File content … exceeds maximum allowed tokens` and nothing else. That is the worst case to lose the map on — the bigger the file, the more the map is worth and the more a full read would have cost.
> 
> Those refusals also land on files the hook's own threshold never sees, because the two count different things: `--min-lines` counts lines, the host's limit counts bytes. A 90-line file of 355 KB trips the host and no threshold of ours.
> 
> Separately, the map itself was never bounded. A minified bundle is one line holding thousands of declarations, so 255 KB of it mapped to 207 KB — roughly twice the tokens the refusal was about, and on the refusal channel that same map replaced a read the agent never paid for.
> 
> ## What
> * **A map for a read the host refuses.** The installer registers the hook under `PostToolUseFailure` too, and a new response channel delivers the map through `hookSpecificOutput.additionalContext`. Verified on a live host: the event fires for a `Read` the host rejected over its size limit, and the context reaches the model.
> * **On that event the line-count and range gates drop.** The read is not happening again, lines are not what refused it, and the host's own error tells the agent to retry with `offset`/`limit` — so gating that retry out would leave the hole open. On the refusal channel both gates stay, where a substitution would replace what the caller asked for.
> * **The map is bounded, shedding detail before cutting anything.** Order follows what each level costs: doc comments and attributes first, which are detail about a declaration, then fields and private items, which are declarations. Measured on Apache Calcite's `SqlFunctions.java` (7771 hand-written lines): 142,692 bytes in full, 74,607 with all 920 declarations once docs and attributes go. Both markers name what is missing and the command that returns it, because a silently short map reads as a file that ends early.
> * **`status` reports `hook partial`** when only some of the events an install writes are registered — the state every existing install is in after this change — with the remedy on stderr, where the repo's contract puts qualifications of a delivered answer.
> * **No empty hook events left behind.** `json_hook::remove` drops an event key once it is empty, so installing or uninstalling no longer leaves `"PreToolUse": []` in a file the user owns.
> * **Internals that make the above possible.** `decide` returns the bare map and `io` owns the framing, so one command serves three events with three notices; `Channel::is_additive()` is the single place that says whether a payload replaces something or is added to it.
> 
> A normal large `Read` behaves as before: `PreToolUse` refuses the call and carries the map — now bounded.
> 
> ## How to verify
> `cargo test` — full suite green, `clippy --all-targets` clean.
> 
> Gates: `failed_ranged_read_is_substituted_but_a_successful_one_is_not` asserts both directions of the range gate, `failed_read_substitutes_below_the_line_threshold` the line gate, `failed_read_still_needs_a_parseable_file` that only those two gates moved. Bounding: `a_large_hand_written_file_sheds_detail_instead_of_declarations` builds a fixture whose full map exceeds the cap and asserts nothing was truncated, the shed is named, and the first and last declarations survive; `a_minified_bundle_is_trimmed_and_points_at_the_whole_map` covers the backstop. Registration: `install_hook_registers_pre_tool_use_and_the_failure_event` also asserts `PostToolUse` is _not_ registered. Status: `status_reports_partial_for_an_install_missing_the_failure_event` and `status_is_not_partial_after_install_completes_it`. Notices: `every_notice_keeps_its_command_on_one_line` catches a wrap inside the backticks that would comment out the symbol argument.
> 
> By hand, on the file the host refuses:
> 
> ```
> $ echo '{"hook_event_name":"PostToolUseFailure","tool_name":"Read","tool_input":{"file_path":"huge.rs"}}' | ast-bro hook --protocol claude-code --min-lines 200
> {"hookSpecificOutput":{"hookEventName":"PostToolUseFailure","additionalContext":"# ast-bro supplies this structure map, with line ranges, because the Read did…"}}
> ```
> 
> Installing over a settings file leaves the user's own hooks untouched; uninstalling leaves no empty keys.
> 
> ## Note on #34
> #34 asks for a channel that reports the substitution as a success rather than a failed tool call. That channel does not exist for the built-in `Read` on Claude Code 2.1.223: `hookSpecificOutput.updatedToolOutput` replaces the result of an MCP tool and is ignored for `Read` and `Bash`. Measured with a hook returning `updatedToolOutput` and `additionalContext` in one object — the second lands every time, the first never does — with workspace trust, field name, and envelope parsing ruled out ([evidence](https://github.com/aeroxy/ast-bro/issues/34#issuecomment-5233506130)).
> 
> So `PreToolUse` remains the registered channel, and `Channel::Replace` stays in the code unregistered, carrying that measurement and reachable from a hand-written `PostToolUse` entry for anyone whose host honors the field. This PR does not close #34.
> 
> Refs #34.
> 

Thanks man. Few suggestions:

**1. `MAX_MAP_BYTES` interaction with markers**
```rust
const MAX_MAP_BYTES: usize = 64 * 1024;
const MARKER_RESERVE: usize = 1024;
```
In the shedding path you check `map.len() <= MAX_MAP_BYTES.saturating_sub(MARKER_RESERVE)` then append a marker that can be a few hundred bytes. In the truncation path you subtract the reserve before cutting. The final payload can therefore be up to ~`MAX_MAP_BYTES + a bit`. The tests already allow `+ 400`. Consider either:
- documenting that the hard ceiling is soft, or
- making the final length check after the marker is added and trimming further if needed.

Minor; not a correctness issue.

**2. `map_levels` order and the empty-shed case**
The first level is the full map (`""`). If it already fits you return it immediately — good. The subsequent levels always emit a marker. That is correct, but the marker text for the second level (“doc comments and attributes”) is accurate only if those were the *only* things dropped; later levels correctly accumulate the list. No change needed, just noting the design is consistent.

**3. `Channel::Replace` is still reachable**
Any hand-written `PostToolUse` entry will hit it. The comment in `event.rs` is clear. Consider a one-line note in the README or `install` help that registering it yourself currently does nothing useful on Claude Code for `Read` (so people don’t waste time).

**4. Augment notice line break**
```text
# ast-bro supplies this structure map, with line ranges, because the Read did
# not return the file.
```
The test deliberately checks for the split. Fine, but a future edit that re-wraps the notice could break the assertion. The “command on one line” test is the more important guard.

**5. Possible future hardening**
- The shim does not inspect the failure reason. If Claude Code ever starts sending structured error info, you could gate more precisely (e.g. only size/token refusals). Current conservative approach is right.
- `cap_map` counts lines with `split_inclusive('\n')`. Fine for the maps you generate; just be aware of files that lack a final newline.

**6. README wording**
The new paragraph about the substitution arriving as a blocked tool call is accurate for the `PreToolUse` path. It might be worth a short sentence clarifying that the *failure* path is additive and does not block anything (the host already failed the call).

vlsi:
Thanks — all six addressed, and the review that followed turned up two bugs in the fix for the first one. Pushed as `34ba60f`.

**1. Soft ceiling** — took the second option and made it hard rather than documenting it as soft. `MARKER_RESERVE` is gone. Both markers are built *before* the fit decision and counted against the budget, and `cap_map` sizes its marker against the widest numbers it can carry, cuts, then rebuilds it with the real ones — never wider, since `kept ≤ map.len()` and `dropped ≤` the line count. `io::MAX_NOTICE_BYTES` (the widest of the three notices plus what `payload()` adds, computed by a `const fn`) is held back too, so the ceiling now bounds what is *delivered*, not just the map inside it. A sweep of 458 payloads across every tracked file plus adversarial fixtures, both channels: largest 65,535, none over. The `+ 400` slack is out of the tests.

**2. `map_levels` order** — no change, as you said.

**3. `Channel::Replace` reachable by hand** — README now says registering a `PostToolUse` entry yourself does nothing useful for `Read` on 2.1.223, for the same reason the installer skips it.

**4. Augment notice wrap** — the test no longer pins where it wraps: it un-wraps the notice and asserts the claim, and the negative half now matches the same un-wrapped string (it was still checking the raw constant, so a re-wrap splitting a forbidden phrase would have passed vacuously). The command-on-one-line test remains the real guard.

**5. Files without a final newline** — covered by `cap_map_handles_a_map_without_a_trailing_newline`.

**6. README on the failure path** — added: nothing is blocked there, the map arrives beside the error. Also documented the 64 KB cap and the shedding order, which the section never mentioned.

**Two bugs the cross-review found in the ceiling work, both worth knowing about:**

- `cap_map` stopped at the first line that did not fit instead of skipping it. One generated declaration with 9000 parameters therefore hid everything after it: a 401-declaration file delivered 957 bytes with **zero** declarations — the exact failure that function's doc comment claims to prevent. Now over-long lines are skipped and counted: 400 of 401.
- Making it skip introduced a worse problem. The map became a non-contiguous selection, and indentation is its only parent-child signal, so a skipped type header left its members filed under the *previous* type — a map asserting a structure the file does not have. The skip is now parent-aware: a dropped line takes every following line of greater indent with it. The regression test computes each line's ancestors from the original map and asserts every kept line still has all of them; with the fix removed it fails naming the exact orphaned pair.

Also fixed from the same round: the trim path had been reporting the already-shed map's size as the file's full map size, and the fallback kept the *last* shedding level rather than the smallest — the `max_members` level trades members for `+N more member(s)` markers and can come out larger than the level before it.

Full suite green, `clippy --all-targets` clean.

vlsi:
Ran the hook end to end against a real host rather than only piping JSON at it. Five nested `claude -p` sessions on Claude Code 2.1.223, each through a wrapper that logged both the hook's stdin and its own reply, and I read the actual tool results out of `--output-format stream-json` instead of taking the model's word for what it saw. Now at `b49d90b`.

| Scenario | Result |
| --- | --- |
| Refusal channel, 40-line file | Hook fired, the model received the map with the notice leading it, and the raw source is absent from the result. `is_error: true` — the framing #34 is about, unavoidable on this channel |
| Failure channel, 90 lines / 355 KB | `PreToolUse` passed it through (under the threshold), the read ran and failed, `PostToolUseFailure` fired, the map arrived. The model: *"Read failed (355.4 KB exceeds 256 KB limit), and yes, I received an ast-bro structure map showing 90 methods; one function is `wide_0()`"* |
| The cap, 367 KB minified bundle the host refused | 65,505 bytes delivered against the 65,536 ceiling; the marker names 13,446 dropped lines and the real full-map size of 260,123 bytes; the model still got usable declarations out of it |
| Ranged retry, `Read(huge.rs, offset=1, limit=2000)` | The read failed again, `PostToolUseFailure` fired, the map arrived. This is the range-gate change working: before it, that gate passed the retry through and the agent got a second bare error |
| Pass-through, one-line file | The model received the real file contents, no map |

## One correction the measurements forced

I had been describing the refusal as "past its size or token limit" — in the README and in three doc comments — with the token half marked unverified. I tried to isolate it: a 201 KB file of 150 lines, under the size limit and well over any 25k-token ceiling. **The host read it successfully**, returning 40 KB for a 201 KB file, so it truncates long lines rather than failing on tokens.

So the reproducible trigger is the file-size limit, and a token-only refusal looks unreachable for a file under 256 KB. The wording is now what was measured: the README says "measured against its file-size limit, 256 KB on Claude Code 2.1.223", and "bytes or tokens" is gone from the comments.

For what it is worth, the first attempt at that isolation failed for a different reason — a 2300-line file was caught by the refusal channel before the read ever ran, which is correct behaviour but tells you nothing about the failure path. Worth mentioning because it is the easy way to conclude the wrong thing here.

Suite green, `clippy --all-targets` clean.

aeroxy:
Nice work @vlsi the live-host run is the part I wanted, and correcting your own "size or token limit" wording against a measurement is the right instinct.

Pulled `b49d90b`: suite green here too, clippy clean. One hole left, and it's in exactly the invariant you added the regression test for.

**1. A blank line resets `skipped_parent`, so members still outlive dropped headers**

`_render_decl` pushes an empty line after *every* `Class | Struct | Interface | Record | Enum | Namespace`, not just top-level ones — so a nested type drops a blank line into the middle of its parent's member list:

```
public class Outer  L1-6
    public static class Inner  L2-4
        void innerMethod()  L3
                                      <- blank; indent computes to 0
    void outerMethodAfterInner()  L5
```

In `cap_map`, `indent = line.len() - line.trim_start_matches(' ').len()` is 0 for `"\n"`, so the blank clears `skipped_parent` and the next member is re-evaluated on its own — kept if it fits, filed under the previous type. `a_trim_never_keeps_a_line_whose_parent_it_dropped` passes because its fixture only puts the blank *after* the last member of each struct.

Probe straight at `cap_map`, wide headers and short members:

```rust
map.push_str(&format!("public class Outer{n}{}  L1\n", "X".repeat(3000)));
map.push_str("    public static class Inner  L2\n");
map.push_str("        void innerMethod()  L3\n");
map.push('\n');
map.push_str(&format!("    void survivor{n}()  L5\n"));
map.push('\n');
```

→ 14 members kept whose class header was dropped.

Reachability is narrow: it needs headers wide relative to their members (long generics, big `implements` lists), because at uniform width the budget simply runs out and nothing past the cut survives anyway. A 1200-class generated Java file — 193 KB full map, 65,505 delivered — gave 0 orphans. So: low blast radius, but it's the exact failure the function's doc comment claims to prevent, and the test reads as covering it.

Fix is to leave the skip state alone on a blank line instead of letting it compute as indent 0, plus a mid-member blank in the fixture so the test actually pins it.

**2. `ALL_HOOK_PATHS` is byte-identical to `INSTALL_HOOK_PATHS`** in all three installers, with the same paragraph of doc comment explaining a divergence that doesn't exist yet. I buy the seam; three copies of it is a lot of prose for a constant that currently differs only in name.

**3. `hook partial` has no opt-out.** Anyone who deliberately deletes the `PostToolUseFailure` entry now gets that stderr note on every `status`, forever, telling them to re-run `install`. Fine if you consider the entry non-negotiable — worth a thought if not.

**4. Old-host compat.** `install` writes `hooks.PostToolUseFailure` unconditionally now. You've verified the new behaviour on 2.1.223; have you checked a version that predates the event ignores the unknown key rather than rejecting the settings file?

Everything else I went after checked out. The budget arithmetic errs the right way — `trim_note` sized against the widest numbers it can carry means the `debug_assert!` can't fire, since `kept ≤ map.len()` and `dropped ≤ total_lines` both hold digit-wise. `MAP_BUDGET` holding back `MAX_NOTICE_BYTES` does make the ceiling bound what's delivered. `prune_empty` stops at the first non-empty ancestor and leaves sibling events alone. And the additive channel drops exactly the two gates you say it drops — `failed_read_still_needs_a_parseable_file` pins the third.

vlsi:
Rebased onto `origin/main` (17 commits) and pushed as `09c411c`. All four addressed, and you were right that the blank line was worth generalising — there were three instances of that class, not one.

## 1. The trim no longer reads structure out of text

Swept 272 files across nine languages before fixing anything, and found two more lines whose indentation does not carry nesting:

- **Separators**, exactly as you described — `_render_decl` pushes one for every `Class | Struct | Interface | Record | Enum | Namespace`, not only top-level ones. **139 of them sit between two indented lines, in 45 of the 272 files**, including this repository's own sources.
- **Doc-comment continuations.** A Javadoc block is stored as one string with embedded newlines, so only its first physical line gets the render prefix and the rest keep the indentation they had in the source. 1,314 lines in the corpus land at columns 1 or 3. Ordinary Calcite Java:

```
    public class ArrayEnumeratorCursor …   ← column 4
        /**                               ← 8
   * Creates an ArrayEnumeratorCursor.    ← 3   ← clears the skip state
        public ArrayEnumeratorCursor(…)   ← 8   ← orphaned
```

Leaving the state alone on a blank line would not have covered that one. So the fix is a level down: `core::render_map_units` now returns the units a map is assembled from — header, one unit per declaration **together with its doc comment**, `+N more member(s)` markers, separators — and `render_map` joins them. `cap_map` consumes units, `unit_indent` reports `None` for a separator instead of column 0, and grouping the doc with its declaration closes the third case, where a comment kept after its declaration was dropped would have read as documentation of whatever followed. Output is byte-identical: 1,088 renders across those 272 files under four flag combinations.

The regression test contains both shapes and checks the claim the function actually makes — every kept unit still has all of its ancestors. Worth noting the first version of it was useless: it derived its expectation from `unit_indent`, the function under test, and passed with the bug restored. It now computes ancestors itself, reports 28 orphans when a separator's column reads as 0, and names them. Your point about narrow reachability held up end to end: on a real Java file built for it, the delivered payload has zero orphans either way, so the unit test is the guard, not the binary.

## 2, 3

Each installer keeps one path constant, and the seam is stated once in `common.rs` where a divergence would actually be needed. The `hook partial` note is now an observation: *"our hook is registered for some of the events `install` writes, not all. `ast-bro install` adds the rest; nothing else is needed if the missing one was removed deliberately."*

## 4 — still unanswered, and my earlier evidence was worthless

Isolating the config breaks auth, so 2.1.50 and 2.1.150 only reach "Not logged in". I checked that with a negative control before believing it: a **deliberately malformed** settings file produces the same message, so settings are not parsed at that point and "it did not complain" established nothing. What remains is an authenticated run of an old version against the real config directory, which can rewrite `~/.claude.json`. Happy to do it behind a backup-and-restore if you want the answer; I did not want to risk it unprompted.

## What the follow-up cross-review found

One thing bigger than anything in your list, a level above where I had been looking. `render_map_for` took the first shedding level that *fit*, never asking what it cost — and shedding and trimming are lossy in different currencies. A 3,000-line file of 2,900 private and 100 public functions delivered **100 declarations in 4,248 bytes against a 65,308-byte budget**, and an all-private file delivered the header and a note: zero. It now builds a candidate per level — whole if it fits, trimmed otherwise — and keeps whichever delivers the most entries, strictly greater so the richer level wins a tie. Those two files now deliver 1,437 and 2,067 declarations; `SqlFunctions.java` went 799 → 803; a small file whose full map fits is still a single render, because the search stops at the first level that fits.

Also from that review: the shed note could glue itself to the header line of a file with no declarations (both paths now share `with_note`), the README described the trim as "line by line" — the model this change replaced — and my `updatedToolOutput` claim was stronger than the measurement. The observation stands (four probes, and the same field did replace an MCP tool's result in the same session, which rules out an envelope or field-name error), but whether the host declines it for every built-in tool or only for one without an output schema is not settled from here; `Channel::Replace` now says that, and `Read` is on the wrong side either way.

Suite green, `clippy --all-targets` clean apart from one pre-existing `no_effect_replace` in `src/search/chunker.rs:1558`, which arrived with the rebase and is not mine to touch here.

aeroxy:
@vlsi Fixed properly, and a level below where I pointed — generalising from "blank lines lie about depth" to "physical lines lie about depth" caught the doc-continuation case, which I would not have found. Moving the trim onto units is the right seam.

Checked the claims rather than taking them:

- **Byte-identical render** — 12 combinations (`src/`, `tests/`, `wiki/` × default / `--detail signatures` / `--no-docs` / `--preset digest`), hashes identical against `a61b8fc`.
- **The regression test actually bites** — I restored the blank-clears-state behaviour by neutering `unit_indent`'s early return, and `a_trim_never_keeps_a_unit_whose_parent_it_dropped` fails. Not vacuous.
- Suite green, 432 lib tests. The one clippy warning is `no_effect_replace` at `src/search/chunker.rs:1558`; confirmed it came in with `edb9653` on main, agreed it is not yours to touch here.
- 2 and 3 done. 4 left open is the right answer, and the negative control on the malformed settings file is exactly why — good instinct not to conclude anything from "it did not complain".

## The other fix in this commit trades the public API away

Ranking candidates by entry count can deliver a payload with none of the file's public surface. 3000 private + 500 public Rust functions, 132 KB full map:

```
payload:  65,504 bytes
private:  1733 kept
public:   0 kept    ← of 500
note:     "map trimmed to 64922 bytes … 1767 entries dropped"
```

The `--no-private` level is **19 KB** — complete, every public symbol, 70% of the budget unspent. It loses 501 to 1734 and gets discarded. Interleaved (each `pub fn` followed by six helpers) it is less stark and still bad: **236 of 500** publics survive.

That is the exact mirror of `a_mostly_private_file_keeps_private_declarations_by_trimming`, which is the tell. Entry count cannot separate "dropped a category, said so" from "dropped the tail, said so", and the tail can be the entire public surface. A two-key ordering — public entries, then total — fixes my file and regresses yours, so it is not the answer either.

The shape that gets both is to stop choosing between whole levels and pack instead: take the fitting level as the floor, then spend what is left of the budget on the shed material until it is full. Mine becomes 500 publics + ~1200 privates, yours becomes 50 publics + ~1400 privates. The budget is a knapsack and the code currently picks one strategy for the whole of it.

## Minor

- `unit_indent` classifies on the *first line* being blank, so a declaration unit whose leading doc line rendered blank would read as a separator — carrying no depth and starting no skipped subtree, i.e. the same bug. I swept 439 files across nine languages and found none, because every adapter keeps the marker (`///`, `/**`) on that line. `unit.trim().is_empty()` rather than `first.trim().is_empty()` makes it structural instead of dependent on that continuing to hold.
- One line in the new README paragraph runs well past the surrounding wrap ("Which one you get is decided by…").

vlsi:
> The budget is a knapsack and the code currently picks one strategy for the whole of it.

PS. I'm currently exploring ladder output in MCP so agent gets multiple output files so it can consult whatever it feels better. CLI output can't easily switch between "output to stdout" and "output to a file" depending on the output size. MCP knows it talks to an agent, so it can switch output shape in the runtime.

aeroxy:
@vlsi like it. it also kills the argument above instead of settling it — if the agent can pull a richer rung itself, picking the wrong one costs a round trip instead of the public API.

two things though.

files re-enter the problem. agent gets a path, calls `Read` on a 200 KB map, host refuses it over the size limit — that's the exact failure this PR closes, one indirection later. plus lifetime: who deletes them, is the one on disk still current, local path in the output. we already answer `resources/list` with `[]` at `src/mcp/mod.rs:93`, that's the seam. rungs get URIs, tool result links them, client fetches what it wants, nothing hits disk. resource support is patchy across clients though, so whatever rung goes inline still has to stand alone with its note — ladder is offered, never leaned on.

and keep it additive. "MCP knows it talks to an agent" is true but it shouldn't change what a call returns based on how big the answer came out. that's the #35 rule already — hint never redirects, `map <dir>` returns what was typed and the cheaper form rides along. stronger here really: a tool whose shape depends on its own output size is one an agent can't reason about before calling it.

doesn't cover the hook either. `PreToolUse`/`PostToolUseFailure` is one-shot, no follow-up call, so `render_map_for` still has to choose right on its own — worth keeping the two separate so the ladder doesn't become a reason to leave the ranking as is.

cheap to build fwiw: `map_levels()` already renders every rung on the way to the decision, so caching by path+mtime and handing out URIs costs about nothing over what the trim spends already. same shape as the graph cache.

separate issue? #50 is fat enough.

vlsi:
Sure MCP will be a separate issue/pr.

> "MCP knows it talks to an agent" is true

Well, Anthropic suggested [code execution a year ago](https://www.anthropic.com/engineering/code-execution-with-mcp) which to me is effectively a bash pipe for MCPs. So "knows it talks to an agent" might not last long though.

vlsi:
Both taken, and the minor again turned out to be the one worth generalising: the fix for it is what made the main one implementable. Pushed as `7e2ceef`.

## 1. The budget is packed rather than chosen

You are right that entry count cannot separate the two kinds of loss, and that a two-key ordering only moves which file it fails on. The shape I landed on is the packing you described, expressed one level down: instead of "take the fitting level as the floor, then spend the rest on the shed material", `cap_map` funds units in the order the ladder itself ranks them — a declaration that is still present at a leaner level is funded before one that is not, and the remaining budget goes to the rest. Order within a rank is the file's own, and so is the order the kept units are written back out in.

Same outcome as your floor-then-fill for the case you measured, and it generalises to the level your description does not reach: the ladder's first step sheds docs and attributes, which removes no declarations at all — it shrinks the units — so "restore the shed material" is a different operation there than at the level below it. Reading the rank off the renders covers every step with one rule, and because the packing happens inside one level, a `+N more member(s)` marker never ends up counting members that were handed back.

Your file, regenerated here as 3000 private and 500 public functions:

| | delivered |
| --- | --- |
| before | 1824 private, **0** of 500 public |
| after | 1348 private, **500** of 500 public |

Worth saying that your own regression fixture was failing the same way — `a_mostly_private_file_keeps_private_declarations_by_trimming` delivered 1824 private and **none** of its 50 publics, and the assertion never looked. It now keeps all 1824 and gains the 50.

The cost, since you will measure it: on a file whose type headers are wide relative to their members, the rank order buys fewer declarations for the same bytes. A generated Java file of 300 classes with 80 methods each went 1897 → 1867 declarations, and on a deliberately extreme fixture (4 KB of generic bounds per header) a reviewer measured 353 → 215, trading 140 private declarations for two public ones and a nested type's members. I take that to be the ladder working as specified — private helpers are the first thing it sheds — rather than a regression, but the number is there if you disagree with the trade.

## 2. The trim no longer reads anything out of the rendered text

`unit.trim().is_empty()` would have closed the case you found. The case behind it is that `cap_map` was still inferring structure from text at all, and this is the third bug of that shape in this PR — blank lines, then doc continuations, then a unit whose first line renders blank. So `render_map_units` now returns `MapUnit { text, kind, parent }`: what nests under what comes off the declaration tree, and each declaration carries a number assigned in pre-order over the *whole* tree, filtered-out declarations included, so the same number names the same declaration in every render. That is also what lets two renders be compared for ranking at all.

`unit_indent` is gone, and with it every text-derived structural decision in the hook — the only remaining text inspection in production code is whether a map ends in a newline before the marker is appended.

Verified byte-identical rendering across 30 combinations (`src`, `tests`, `wiki`, `README.md`, `src/adapters` × default / `--detail signatures` / `--no-docs` / `--preset digest` / `--json` / `--detail names --no-private --max-members 3`).

## 3. README

Re-wrapped, and one more ragged line next to it ("That event cannot replace a result, only / add context"). The paragraph now also says how the budget is spent.

## 4

Still open, on the same terms as last round.

## What the cross-review found in the above

Four items, one of them code.

**The marker overcounted by one byte.** The new fill charged every kept unit a trailing newline, but the join gives the last unit none, so "map trimmed to N bytes" reported one more byte than the payload carried. A unit now pays for a newline only when it follows another, and the byte the marker sits on comes off the budget once. A test asserts the marker's number equals the bytes above it — for a normal trim, for a trim where every declaration is wider than the whole budget so only the header survives, and through `decide` with a shed label in the same marker.

**Two coverage holes in what I had just written.** The orphan test ran `cap_map` with no leaner render, so every rank was equal and the ranked path — the thing the commit adds — was never exercised against a nested tree; nothing asserted the kept units come back in the file's order either. It now runs both funding orders over the same fixture and requires the ranked one to keep strictly more headers, so the second case cannot quietly degenerate into the first. And id stability was pinned only for the visibility filter, while the ids that decide the top rank come from the member cap; that branch is now sampled below, at, and above the member count, the middle case being the one that must not cut. Both fail if the arithmetic they guard is reverted.

**The README over-promised.** It said the public surface "arrives whole", which does not hold when the public surface alone exceeds the cap. On the multiple, the README was right and my test comment was wrong: the fixture's full map is 135,114 bytes against a 65,308-byte budget, so "twice the budget" stands and "three times" is corrected.

## How to verify

`cargo test` — full suite green, 436 lib tests; `clippy --all-targets` clean apart from the `no_effect_replace` in `src/search/chunker.rs` that came in with `edb9653` on main.

Ceiling re-swept after the byte-accounting change: 466 payloads across every tracked file plus adversarial fixtures, both channels, largest 65,535 against 65,536, none over.

aeroxy:
@vlsi this is better than what I asked for. funding by ladder rank and writing back in file order is the packing idea without having to merge two renders, and stable pre-order ids are what make ranking across renders possible at all. also retiring `unit_indent` kills the blank-first-doc-line thing structurally instead of patching the predicate — nicer than the fix I suggested.

checked it rather than trusting it:

- my repro is dead. 3000 private + 500 public → **500/500 publics + 1266 privates**, 65,505 bytes. interleaved variant same, 500/500 + 1201. was 0 and 236 before.
- render still byte-identical — 18 combos (`src/`, `tests/`, `wiki/` × default / signatures / no-docs / digest / `--max-members 3` / `--no-private --no-fields`) against main, across both commits and the whole `MapUnit` refactor.
- suite green, 436 lib tests, clippy same pre-existing `chunker.rs:1558`.
- perf is a non-issue: release build, 18–52 ms on 200–750 KB files, four renders and four trims included.

two things that are load-bearing and only implied right now, worth a line each in the code:

the sort's parent-before-child assumption holds for a reason, not by luck. `_render_decl` returns early for an ineligible declaration, so a child only appears in a render where every ancestor did — rank(parent) ≥ rank(child) always, and equal ranks keep file order under the stable sort. the whole orphan rule sits on that.

and the winner can't omit what a leaner fitting candidate would keep: a rank-2 unit renders identically at L1 and L2, so if L2 fits whole then L1-trimmed contains all of it, and L0's units are never smaller than L1's so L0-trimmed can only win on a tie, which forces the same set. that's the property I was actually after and it now holds.

one nit. the hard ceiling holds on the shed path by accident. `cap_map` reserves the byte `with_note` inserts (`.saturating_sub(1)`), the fits branch checks `map_bytes + note <= MAP_BUDGET` and then `with_note` can add a newline. I landed a file exactly on the boundary — delivered 65,308, i.e. MAP_BUDGET on the nose, and the only reason it isn't 65,309 is that `_render_decl` always emits a trailing separator after a top-level declaration so the join already ends in `\n`. one byte, unreachable today, but "the ceiling is hard" is a claim you swept 458 payloads for. a `debug_assert` there (or the same `saturating_sub(1)`) pins it to construction instead of to a renderer detail.

re code execution — fair, and it cuts the same way. if the MCP surface turns into a bash pipe then "shape switches at runtime because it knows it's an agent" gets weaker, not stronger, and a ladder of resources survives that fine while size-dependent output shape doesn't. one more reason to keep it additive.

vlsi:
Both "worth a line each" turned out to need code rather than a comment, and the nit was the smaller of the two. Pushed as `23bc918` (three commits: `403c078`, `a102e46`, `23bc918`).

## 1. Parent before child — comment, as asked

Stated where the orphan rule depends on it: `_render_decl` renders nothing for a declaration the options exclude, so a child reaches a render only where every one of its ancestors did; a parent's rank is therefore never lower than its members', and the stable sort leaves the two in file order when the ranks are equal. The inline note at the sort is now one line pointing at that, rather than half the argument in each place.

## 2. The required-set property did not hold

Your argument is right up to the point where the trim *skips* rather than stops. A trim's budget is smaller than the fitting level's by the width of the marker it has to carry, so it can run out inside the top rank — and when it does, it skips the one unit too wide for what is left and keeps funding cheaper declarations a rank below. It then out-counts the whole candidate while missing part of the public surface that one carried.

Built the fixture rather than argued about it: one type whose public surface is grown to the largest that still fits whole, its last member wide, private methods behind it cheap.

| ranking | map | publics | privates | shape |
| --- | --- | --- | --- | --- |
| by declarations delivered | 65,284 B | 1335 of 1336 — the wide one gone | 8 | trimmed |
| by required, then declarations | 65,277 B | **1336 of 1336** | 0 | whole, shed note |

The trimmed payload wins the count 1343 to 1336, which is exactly how it wins while being the worse answer. A candidate is now scored `(declarations a whole leaner candidate would also carry, declarations)`, so that trade cannot be won. Your repro is unaffected — still 500 of 500 publics — and `a_mostly_private_file_keeps_private_declarations_by_trimming` still keeps its 1824 helpers, because when the trim *can* fund the whole required set the first key ties and the count decides as before.

The test finds the boundary by search rather than hard-coding it, so it stays on the boundary as the notes change width. With the first key removed it fails naming the missing declaration.

## 3. The ceiling

Fixed by construction rather than by reserving a byte: the fit is decided on the assembled payload, so nothing about how the renderer ended its last unit can decide it. And the claim is now checked where it is finally made — `io::payload` is the last thing that adds to a map, so it asserts the total there instead of leaving it inferred from the two sizings that lead in. A test frames a budget-filling map with each of the three notices and requires it to land on the ceiling exactly, which pins `MAX_NOTICE_BYTES` in both directions: short and the ceiling breaks, wide and the map is smaller than it needed to be.

## What the cross-review found in the above

Three rounds, and the middle one is the interesting one.

**Round 1 — the exact check cost what it was worth.** Measuring on the assembled payload made every shed level assemble one, including the levels that overflow and are discarded: on a minified bundle, four joins of an over-ceiling map plus six more inside `required_ids`. Now `whole_payload` adds the parts up first — a payload is never smaller than its map and its note — and assembles only when that bound leaves it deliverable. `required_ids` is hoisted out of the loop as well; its result cannot differ between levels, since the loop stops at the first level that fits, so a scan from any earlier one finds that same level. Also `SEPARATOR_BYTES`, because the bare `+ 3` lost its explanation while I was mutating that constant to check a test's teeth, and putting the prose back would have left the number typed in two places.

**Round 2 — my own evidence was wrong.** I had claimed the boundary sweep exercises the case where the cheap bound says "maybe" and the assembled payload says "no". It cannot: `_render_decl` emits a separator after every top-level declaration, so every map a renderer produces already ends in a newline and the two measures agree exactly. The exact re-check was therefore unreachable on every fixture in the suite — replacing it with `Some(out)` passed all 440 tests. That is the same accident your nit named, one level along: the byte was invisible because of a renderer detail, and so was the check for it. A unit test now builds the shape the renderer does not produce — units ending in a declaration, sized so the parts add up to exactly the budget — and pins three points: on the nose is refused, one byte under fits and spends the budget exactly, and pushing the separator back on makes the same size deliverable again.

**Round 3 — clean from both reviewers.** Worth flagging that two findings in round 2 were a stale re-read: the reviewer had been pointed at the pre-fix commit and re-derived its own earlier findings from it, with remedies that describe verbatim what the fix already did. Rejected with the file and line of each fix; re-run against the current state it returned nothing.

## How to verify

`cargo test` — 440 lib tests green; `clippy --all-targets` clean apart from the `no_effect_replace` in `src/search/chunker.rs` that came in with `edb9653` on main.

Ceiling re-swept after each change: 472 payloads across every tracked file plus adversarial fixtures, both channels, largest 65,535 against 65,536, none over.

`src/core.rs` is untouched by all three commits, so the byte-identical rendering you checked over 18 combinations still holds.

aeroxy:
@vlsi all clean this round, and you were right that my required-set argument was wrong — I'd assumed the trim and the fitting level spend the same budget, and they don't: the trim is short by the marker's width, which is exactly the room it needs to run out inside the top rank. Good counterexample.

verified rather than trusted:

- 440 lib tests green, clippy clean apart from the known `chunker.rs:1558`.
- all three new guards bite. mutated each: dropped the exact ceiling re-check → `a_map_is_measured_with_the_line_its_marker_takes` fails; dropped the `required` key from `score()` → `a_trimmed_candidate_never_omits_what_a_whole_one_carries` (and the boundary sweep, since the whole candidate then never wins); `SEPARATOR_BYTES` 3 → 2 → `a_full_budget_map_lands_exactly_on_the_ceiling`. that also confirms your round-2 correction — the exact check really was unreachable before the new fixture.
- swept the ceiling wider than your 472: both channels over every file in the repo, `wiki/`, fixtures and my adversarial set at `--min-lines 1` — **5053 payloads, largest 65,535 against 65,536, none over**.
- my repros unchanged, 500/500 publics.
- `src/core.rs` untouched, and I re-ran the 18-combo render check against main anyway: still byte-identical.
- the cheap bound made it faster, not slower: `Gen.java` 47 ms → 18 ms, bundle unchanged at ~20 ms.

and the scoring is right by construction now, not by fixture: the levels' declaration sets are nested decreasing, `required_ids` returns the richest *fitting* leaner level's set — a superset of every leaner whole answer's — so a trimmed candidate can only win by tying that key, and can't omit anything a whole candidate carried. the `fits` break is safe for the same reason: any level past the first fitting one is a subset on both keys.

I pushed the constants nit myself as `afbba4c` since you're out — didn't want it sitting on the branch for a week. all three now live in `io` rather than two of them in `decide` reaching across for the third, which follows your own reasoning for putting the assertion in `payload`: that's where the ceiling is either kept or broken. `MAX_NOTICE_BYTES` went private, `decide` just reads `io::MAP_BUDGET`, so the dependency runs one way. no behaviour change — same values, 440 green, same `cargo doc` warnings, and I spot-checked payload sizes (65,535 on the interleaved fixture, 65,505 on the 829 KB bundle). revert it if you'd rather it went in a `budget` module, no argument from me.

nothing blocking. enjoy the leave.

Issue #31, opened 2026-07-28
callers: a same-name method in the calling file shadows an explicit receiver and silently drops the rest of the chain


vlsi:
## Summary

`sb callers` silently under-reports Java call sites when the calling file happens to declare a method whose name matches the head of a call chain. The same-file resolution pass binds the call to the local method and ignores the explicit receiver, tagging that wrong edge `Exact`. Everything after the head of the chain then degrades to `Ambiguous`, and `callers` drops ambiguous edges without printing them or warning.

The result is a blast-radius answer that looks complete and is not. I used `sb callers` to price an API rename and got a number that was 33% low, with nothing in the output to suggest the tool was unsure.

Version: `ast-bro 3.0.0` (Homebrew), macOS 15.5 arm64.

## Minimal reproduction

Three files in `com.example.api`:

```java
// src/main/java/com/example/api/Prefs.java
package com.example.api;

public class Prefs {
  public boolean forDate() {
    return true;
  }
}
```

```java
// src/main/java/com/example/api/Ctx.java
package com.example.api;

public class Ctx {
  private final Prefs prefs = new Prefs();

  public Prefs getPrefs() {
    return prefs;
  }
}
```

```java
// src/main/java/com/example/api/Conn.java
package com.example.api;

public class Conn {
  private final Ctx ctx = new Ctx();

  public Ctx getCtx() {
    return ctx;
  }
}
```

And a caller in a different package that declares its own `getCtx()`:

```java
// src/main/java/com/example/impl/Caller.java
package com.example.impl;

import com.example.api.Conn;
import com.example.api.Ctx;

public class Caller {
  private final Conn connection = new Conn();

  /** Same name as Conn.getCtx(), declared in this very file. */
  protected Ctx getCtx() {
    return connection.getCtx();
  }

  /** Chain whose head shares a name with the local method above. */
  public boolean viaChain() {
    return connection.getCtx().getPrefs().forDate();
  }
}
```

`git init && git add -A && git commit`, then:

```bash
sb callees viaChain
```

Actual:

```text
# 3 callee(s) for '…::Caller::viaChain':
…/Caller.java:16    …/Caller.java::com.example.impl::Caller::getCtx  (Exact)
…/Caller.java:16    [unresolved] forDate  (Ambiguous)
…/Caller.java:16    …/Ctx.java::com.example.api::Ctx::getPrefs  (Inferred)
```

[...]

PR #52, opened 2026-08-09
feat: report a truncated frontier on deps, reverse-deps, impact, and trace

### LINKED ISSUE #32: Report truncation when a result cap is hit

vlsi:
## Why

`--depth` bounds six walks in `ast-bro`, and until now only two of them said when it bit. The other four ended the same way whether they ran out of graph or ran out of depth, so a partial answer was indistinguishable from a complete one:

```console
$ sb reverse-deps src/core.rs --depth 1
src/core.rs ← imported by: (33 total)
$ sb reverse-deps src/core.rs --depth 5
src/core.rs ← imported by: (54 total)
```

Nothing in the first output says 33 is a prefix of 54. That is proposal 3 of #32, the last part still open — proposals 1 and 2 (the `--limit` totals and the JSON `total` / `truncated` fields) shipped in #31-#37, and `callers` / `callees` already carry `frontier_truncated`.

`trace` had the sharper version of the same problem: it answered every unreachable target with "the chain likely breaks at a dynamic-dispatch / framework boundary", including when the BFS had simply hit `--depth 12` with callees left to follow.

## What

Every depth-bounded command now reports whether the walk stopped or the graph did.

| Command | JSON | Text |
| --- | --- | --- |
| `deps` | `frontier_truncated` | stderr note |
| `reverse-deps` | `frontier_truncated` beside `total` / `truncated` | stdout header + stderr note |
| `impact` | `frontier_truncated` per report | stderr note |
| `trace` | `frontier_truncated` on both branches | a different diagnosis, not the dispatch one |

- **`deps` / `reverse-deps`** — new `forward_info` / `reverse_info` return a `DepTraversal`; the plain wrappers stay for callers that only want hits. `reverse-deps` also qualifies the count on stdout — `(33 total within --depth 1)` — because that number is what a reader quotes. The qualifier follows the noun, so a `(\d+) total` matcher keeps working on the capped form.
- **`impact`** — the flag sits on the report, not on a section: the four direct sections are pinned to depth 1 whatever `--depth` says, so only the transitive cone and the counts derived from it can be cut short. Both walks feed it, including the multi-source construction-site walk that type targets take.
- **`trace`** — a capped search now says `no static call path found within --depth 1` and names the flag to raise, instead of asserting a dispatch boundary the evidence doesn't support.

Two rules govern when the flag is set, and both exist to keep the raise-`--depth` hint honest:

- **Only an edge into a node the walk never visited counts.** An edge back into visited territory is not more to see, and would make the hint a lie on any cyclic import graph.
- **Predicate-rejected edges still count.** `--exclude-tests` filters what gets reported, not what gets expanded, so a qualifying node behind an excluded one is genuinely reachable by raising `--depth`.

`impact` clears the flag where raising `--depth` could not change the answer: `--mode deps` and `--mode dependents`, whose report is built entirely from depth-1 sections, and `--exclude-tests` when the affected-tests list was the only consumer — that list is hard-coded to zero at every depth. It deliberately *keeps* the flag on a zero `transitive_count` under `--mode all`, since a zero produced by a depth cap is exactly the confidently-wrong answer #32 is about.

The two note wordings now live in one helper each, which also removes the duplicate copy of the `callers` note that the CLI and MCP had been maintaining separately. `deps::traverse::bfs` loses its `limit` parameter: every call site already passed `UNLIMITED` and display trimming happens above the traversal, so the parameter only created a path where the walk stopped without reaching a cutoff node and still reported `frontier_truncated: false`.

### Behavior notes

- `callers` / `callees` / `impact` print no note at `--depth 1`, where "the callers have callers" is the norm rather than a qualification. `deps` / `reverse-deps` default to 3, so theirs fires at every depth. The JSON flag is set regardless.
- MCP has one channel, so the note is prepended to the response body there.
- `ast-bro.deps.v1`, `ast-bro.reverse-deps.v1`, `ast-bro.impact.v1`, and `ast-bro.trace.v1` gain a field; no field changed meaning or went away.

## How to verify

```bash
cargo test --release
```

664 tests, clippy clean. 24 new assertions across `tests/deps_e2e.rs`, `tests/impact_e2e.rs`, `tests/calls_e2e.rs`, `tests/mcp_e2e.rs`, and `src/calls/trace.rs` — including the two cases where the flag must *not* fire (a depth cap that coincides with the end of the graph, and an edge back into a visited node) and all four `reverse-deps` header shapes.

Manual check of the original repro:

```bash
sb reverse-deps src/core.rs --depth 1 --json --compact
```

Reviewed by Codex and a Claude subagent over three rounds; seven findings fixed, one rejected on the mechanism that `bfs` expands through predicate-rejected nodes (pinned as `impact_all_mode_keeps_the_flag_on_a_zero_transitive_count`).

Closes #32


aeroxy:
> ## Why
> `--depth` bounds six walks in `ast-bro`, and until now only two of them said when it bit. The other four ended the same way whether they ran out of graph or ran out of depth, so a partial answer was indistinguishable from a complete one:
> 
> ```
> $ sb reverse-deps src/core.rs --depth 1
> src/core.rs ← imported by: (33 total)
> $ sb reverse-deps src/core.rs --depth 5
> src/core.rs ← imported by: (54 total)
> ```
> 
> Nothing in the first output says 33 is a prefix of 54. That is proposal 3 of #32, the last part still open — proposals 1 and 2 (the `--limit` totals and the JSON `total` / `truncated` fields) shipped in #31-#37, and `callers` / `callees` already carry `frontier_truncated`.
> 
> `trace` had the sharper version of the same problem: it answered every unreachable target with "the chain likely breaks at a dynamic-dispatch / framework boundary", including when the BFS had simply hit `--depth 12` with callees left to follow.
> 
> ## What
> Every depth-bounded command now reports whether the walk stopped or the graph did.
> 
> Command	JSON	Text
> `deps`	`frontier_truncated`	stderr note
> `reverse-deps`	`frontier_truncated` beside `total` / `truncated`	stdout header + stderr note
> `impact`	`frontier_truncated` per report	stderr note
> `trace`	`frontier_truncated` on both branches	a different diagnosis, not the dispatch one
> * **`deps` / `reverse-deps`** — new `forward_info` / `reverse_info` return a `DepTraversal`; the plain wrappers stay for callers that only want hits. `reverse-deps` also qualifies the count on stdout — `(33 total within --depth 1)` — because that number is what a reader quotes. The qualifier follows the noun, so a `(\d+) total` matcher keeps working on the capped form.
> * **`impact`** — the flag sits on the report, not on a section: the four direct sections are pinned to depth 1 whatever `--depth` says, so only the transitive cone and the counts derived from it can be cut short. Both walks feed it, including the multi-source construction-site walk that type targets take.
> * **`trace`** — a capped search now says `no static call path found within --depth 1` and names the flag to raise, instead of asserting a dispatch boundary the evidence doesn't support.
> 
> Two rules govern when the flag is set, and both exist to keep the raise-`--depth` hint honest:
> 
> * **Only an edge into a node the walk never visited counts.** An edge back into visited territory is not more to see, and would make the hint a lie on any cyclic import graph.
> * **Predicate-rejected edges still count.** `--exclude-tests` filters what gets reported, not what gets expanded, so a qualifying node behind an excluded one is genuinely reachable by raising `--depth`.
> 
> `impact` clears the flag where raising `--depth` could not change the answer: `--mode deps` and `--mode dependents`, whose report is built entirely from depth-1 sections, and `--exclude-tests` when the affected-tests list was the only consumer — that list is hard-coded to zero at every depth. It deliberately _keeps_ the flag on a zero `transitive_count` under `--mode all`, since a zero produced by a depth cap is exactly the confidently-wrong answer #32 is about.
> 
> The two note wordings now live in one helper each, which also removes the duplicate copy of the `callers` note that the CLI and MCP had been maintaining separately. `deps::traverse::bfs` loses its `limit` parameter: every call site already passed `UNLIMITED` and display trimming happens above the traversal, so the parameter only created a path where the walk stopped without reaching a cutoff node and still reported `frontier_truncated: false`.
> 
> ### Behavior notes
> * `callers` / `callees` / `impact` print no note at `--depth 1`, where "the callers have callers" is the norm rather than a qualification. `deps` / `reverse-deps` default to 3, so theirs fires at every depth. The JSON flag is set regardless.
> * MCP has one channel, so the note is prepended to the response body there.
> * `ast-bro.deps.v1`, `ast-bro.reverse-deps.v1`, `ast-bro.impact.v1`, and `ast-bro.trace.v1` gain a field; no field changed meaning or went away.
> 
> ## How to verify
> ```shell
> cargo test --release
> ```
> 
> 664 tests, clippy clean. 24 new assertions across `tests/deps_e2e.rs`, `tests/impact_e2e.rs`, `tests/calls_e2e.rs`, `tests/mcp_e2e.rs`, and `src/calls/trace.rs` — including the two cases where the flag must _not_ fire (a depth cap that coincides with the end of the graph, and an edge back into a visited node) and all four `reverse-deps` header shapes.
> 
> Manual check of the original repro:
> 
> ```shell
> sb reverse-deps src/core.rs --depth 1 --json --compact
> ```
> 
> Reviewed by Codex and a Claude subagent over three rounds; seven findings fixed, one rejected on the mechanism that `bfs` expands through predicate-rejected nodes (pinned as `impact_all_mode_keeps_the_flag_on_a_zero_transitive_count`).
> 
> Closes #32
> 

### Verified

**The repro is fixed, and the fix is the right shape.**

```
$ sb reverse-deps src/core.rs --depth 1
# note: reverse-deps stopped at --depth 1 with unvisited files beyond it; raise --depth to walk further
src/core.rs ← imported by: (33 total within --depth 1)

$ sb reverse-deps src/core.rs --depth 5
src/core.rs ← imported by: (54 total)
```

All four header shapes are right, and the compat claim holds — `(\d+) total` still matches every one, including `(33 total within --depth 1; showing 5 — raise --limit to see the rest)`.

**`trace`'s two diagnoses are genuinely different**, which was the sharper half of the bug:

```
--depth 1   no static call path found within --depth 1
            the walk stopped at the depth cap with unexplored callees beyond it;
            raise --depth before concluding the chain breaks
default     no static call path found
            the chain likely breaks at a dynamic-dispatch / framework boundary
```

**The deps mid-walk check is correct even though it differs from its two siblings.** `calls::traverse` and `trace` both defer the cutoff check until `seen`/`visited` is final and say why; `deps::bfs` checks inline. That looked like the bug until I worked it through: `deps::bfs` is a plain FIFO over uniform edges, so every node at depth ≤ `max_depth` is already in `seen` by the time the first cutoff node pops, and `seen` never grows afterward. The asymmetry is justified — `callers_multi` has seeds at mixed depths, which is exactly what breaks the FIFO argument. Worth a sentence in `deps/traverse.rs` saying so, since the sibling comments assert the opposite rule.

**`impact`'s gate tracks section presence, not mode names.** Checked the four direct sections really are pinned to depth 1, and that `test_calls` comes off the depth-bounded walk, so `&= All || tests_reported` clears the flag exactly where the report has nothing the walk produced. Keeping it on a zero `transitive_count` under `--mode all` is the right call for the reason given.

**Dropping `limit` from `deps::traverse::bfs` is behaviour-preserving** — all three call sites passed `UNLIMITED`. And `trace` inserts its seeds into `visited`, so an edge back into a seed can't produce a false positive.

**The note isn't noisy at the default depth.** Measured over 80 files: `deps` fires on 3, `reverse-deps` on 12. The "an import cone past three hops is a fact about the file" argument holds — this is the one design choice I expected to push back on and the data doesn't support it.

### Finding: the `limit` trap is still live in `calls::traverse::bfs`

The commit message justifies removing the parameter from the deps walk:

> the parameter only created a path where the walk stopped without reaching a cutoff node and still reported `frontier_truncated: false`

That path is untouched in `calls::traverse::bfs`, which still takes `limit` and still early-returns at `out.len() >= limit` (lines 358, 411, 427) with `frontier_truncated` at whatever it was — necessarily `false`, since the return happens during expansion, before any cutoff node pops.

And unlike deps, it isn't a dead parameter. Three live call sites pass finite limits:

```
src/context.rs:289   traverse::callers(calls, &c.qn, 1, 50, …)
src/context.rs:369   traverse::callers(calls, &c.qn, 2, 50, …)
src/context.rs:567   traverse::callers(calls, method_qn, 1, 20, …)
```

A symbol with more than 50 direct callers hits that return, and the computed flag is wrong. Nothing is user-visible today only because `callers()` throws the flag away — `context` has no `--depth` and reports no frontier. So this is a wrong value computed on a live path that happens to be discarded, one `callers_info` call away from being surfaced, and it now contradicts the invariant this PR just established next door.

Cheapest fix that keeps the shape: have the limit path set `frontier_truncated = true` before returning, since stopping early on a result count is by definition "there was more". Or move `context`'s trimming above the traversal, the way this PR moved `reverse-deps`'.

### Nit: the wiki's numbers don't reproduce

`wiki/deps.md` says the flag exists because `reverse-deps --depth 1` "came to report `32 total` for a file with 53 importers". No file under `src/` yields either number, and `src/core.rs` — the file the PR description's own repro uses — gives 33 and 54. Off by one on both, so it reads like a measurement taken a commit or two earlier. Small, but it's a measured claim in the paragraph explaining why an unqualified count is dangerous.

PR #56, opened 2026-08-10
docs: note that the CLI and MCP surfaces carry one feature set

vlsi:
## Why

#49 documented 62 CLI arguments, and then had to document 21 MCP properties in a second round, because the guard it added walks `Cli::command()` and can never reach the hand-written JSON literal in `src/mcp/tools.rs`. Both surfaces are now guarded, but only for *descriptions* — the tests check that an existing argument describes itself, not that it exists on both sides.

The argument sets themselves have already drifted. Comparing a live `tools/list` against `--help` on each subcommand:

| Tool | On the CLI | Missing from MCP |
| --- | --- | --- |
| `digest` | `--detail`, `--no-private` / `-fields` / `-docs` / `-attrs` / `-lines`, `--preset` | the whole three-axis form from #37 — the MCP schema still offers only `include_private` / `include_fields` |
| `map` | `--preset`, `--include-private`, `--include-fields` | — |
| `callers` | `--tests`, `--exclude-tests` | — |
| `reverse-deps` | `--tests`, `--exclude-tests` | — |
| `search`, `context`, `impact` | `--rebuild` | — |
| `show` | `<TARGET>...`, several targets | `path` is a single string |

Drift runs the other way too: the `rebuild` property #49 removed from `callers` / `callees` / `trace` was in the schema but absent from the args struct, so serde dropped it and the tool advertised a lever it did not have.

Defaults happen to agree everywhere I checked — `depth` 1 / 2 / 3 / 12, `limit` 200, `max_members` 50, `budget` 8000, `top_k` 10, `show --limit` 20 — but nothing holds them there.

Nothing in the repository said the two surfaces are meant to match, so this records it where a change to either one starts.

## What

One paragraph appended to `AGENTS.md` (which `CLAUDE.md` and `GEMINI.md` symlink to): the tools mirror the CLI subcommands, the six maintenance commands are deliberately CLI-only, the existing guards do not cover argument *existence*, and the two declarations live in different files, so they change in the same commit.

Documentation only — no code, no tests, no behavior change.

## How to verify

Reproduce the table above against a build of `main`:

```bash
printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{}}' '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}' | ast-bro mcp
```

## Follow-up

The note is a convention, not a guard. The mechanical version is a `cli_mcp_parity` test that compares the argument names of `Cli::command()` against `tools::list()` and keeps an explicit allowlist of intended differences (`--compact`, the six maintenance commands, and `show`'s single target — if that one is a decision rather than an omission). Happy to open an issue for that plus the seven gaps above, if you want them tracked separately.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,3 +18,7 @@ The architecture page links to eight deeper wiki files — read them on-demand w
 - [wiki/squeeze.md](wiki/squeeze.md) — log/text token compression (squeeze): multi-stage pipeline, reversible legend, degenerate fallback
 - [wiki/network-security.md](wiki/network-security.md) — model download, TLS policy, mirror fallback
 - [wiki/file-filtering.md](wiki/file-filtering.md) — what gets walked, ignore layers, escape hatches, shebang detection, test-file heuristics
+
+## CLI and MCP are one feature set
+
+Each MCP tool in `src/mcp/tools.rs` mirrors a CLI subcommand; only the maintenance commands (`install`, `uninstall`, `status`, `hook`, `prompt`, `mcp`) are CLI-only. A flag added, renamed, or redefaulted on one surface is unfinished until the other carries it — the documentation guards check that an existing argument describes itself, not that it exists on both sides. The two are declared in different places: clap derives in `src/lib.rs`, a hand-written JSON literal in `src/mcp/tools.rs`. Change them in the same commit, and state any intended asymmetry in the tool description.


aeroxy:
Thanks @vlsi — the convention is worth writing down, and AGENTS.md is the right place for it, since that's where a change to either surface starts.

Both verifiable claims hold. 25 clap subcommands minus 19 MCP tools is exactly the six you name as CLI-only, and `index` is correctly not among them. Locations check out too — `src/lib.rs:48` and `src/mcp/tools.rs:16`.

Two things before this merges.

**The guards aren't on `main` yet.** "the documentation guards check that an existing argument describes itself" is true of #49, which is still open — nothing on `main` walks `Cli::command()` for help text. If this lands first, AGENTS.md describes tests that don't exist. Either merge #49 first, or attribute it inline ("the description guards in #49 check …") so the sentence is true in both orders.

**The rule ships with no exceptions, and `main` already breaks it twice.** "state any intended asymmetry in the tool description" is the right hatch, but nothing has used it:

- `--compact` is on every CLI subcommand that takes `--json`, and appears zero times in `src/mcp/tools.rs`. No tool description mentions it.
- `show` takes `<TARGET>...` on the CLI and a single `path` string on MCP.

As written, someone arriving fresh reads the paragraph as a claim that the two surfaces currently match. One clause naming those two as the known asymmetries turns it into a convention with a starting point instead.

Minor: MCP names underscore where the CLI hyphenates (`reverse_deps`, `find_related`). "mirrors a CLI subcommand" glosses that; half a sentence saves the next person a grep.

On the follow-up — yes, please open the issue for `cli_mcp_parity` plus the gaps. I confirmed the drift you list: `digest` on MCP still offers only `include_private` / `include_fields`, `callers` and `reverse_deps` have no `tests` / `exclude_tests` while `impact` does, and `search` / `context` / `impact` have no `rebuild`. One correction to the description above: `rebuild` on `callers` / `callees` / `trace` is wired on `main` (`src/mcp/tools.rs:865-891`) — it's a phantom only after #49, not today.

And #57 mechanizes the "redefaulted" third of this rule, so whichever of the two lands second should cross-reference the other.

PR #57, opened 2026-08-10
refactor: read CLI and MCP defaults from one set of constants

vlsi:
## Why

Every default was declared three times, in three languages, with nothing tying them together:

```rust
#[arg(long, default_value_t = 200)]                                  // clap, src/lib.rs
#[serde(default = "default_two_hundred")]                            // serde, src/mcp/tools.rs
"description": "Cap result count (default 200)."                     // prose, same file
```

Retune one and the other two keep describing the old behavior. The third copy is the dangerous one: an agent reading the schema has no way to check the number it is told against the number the tool applies.

The helper names made it worse. `default_one` served both `callers.depth` and `callees.depth`, so those two defaults were coupled by an implementation detail. Naming by value is how that happens; `default_two_hundred` and `min_lines: 200` look interchangeable and are not.

## What

`src/defaults.rs` holds each value once, named for what it bounds — `LIMIT`, `CALL_DEPTH`, `FILE_DEPTH`, `IMPACT_DEPTH`, `TRACE_DEPTH`, `TOP_K`, `BUDGET`, `MIN_SIZE`, `MAX_MEMBERS`, `MAX_DOC_LINES`, `MAX_HEADING_DEPTH`, `SURFACE_MAX_DEPTH`, `SHOW_LIMIT`, `IMPACT_MODE`, `ROOT`, `HOOK_MIN_LINES`. clap takes the constant, serde takes a thin function around it, and the schema interpolates it into the description:

```rust
"description": format!("Cap result count (default {}).", crate::defaults::LIMIT),
```

Migrated: 30 clap defaults across 19 subcommands; 20 serde helpers in `src/mcp/tools.rs`, `src/impact.rs`, `src/context.rs`, and `src/search/mcp.rs` (all four had grown their own `default_dot` / `default_path`); 29 schema descriptions; and the values assembled in bodies rather than declared in attributes — `DigestOptions`, `MapOptions`, `DepOptions`, `SurfaceOptions`, `InstallOpts`, and the `--preset digest` member cap.

That last group is the one worth reading the diff for. It is invisible to both attribute forms, and it is where the drift already was: `run_map_digest` builds `DigestOptions` field by field, so the CLI capped digest members at its own literal 50 while the MCP tool followed the constant. `max_doc_lines` and `max_heading_depth` had four copies each across `core.rs`, `lib.rs`, and `mcp/tools.rs`.

## Guards

Six tests over every `.rs` under `src/`, so the next hard-coded default fails the build:

| Guard | Rule |
| --- | --- |
| `clap_defaults_come_from_this_module` | `default_value` / `default_value_t` names a constant |
| `serde_defaults_come_from_this_module` | `#[serde(default = "…")]` names a helper here |
| `schema_descriptions_interpolate_their_default` | wherever `default` introduces a value in a description, that value is the `{}` a `format!` fills |
| `struct_defaults_come_from_this_module` | a field name that reads a constant anywhere reads it everywhere |
| `impl_default_bodies_hold_no_literals` | an `impl Default` body holds no bare values, whatever the field is called |
| `field_splitter_reads_one_line_bodies` | the textual splitter the two struct guards share |

The last two exist because the derived rule cannot see a name's *first* site: revert `core.rs` and `max_members_per_type` stops reading a constant anywhere, so it drops out of the very set that would have caught it. Asking what an `impl Default` body contains needs no name.

One residual is known and deliberate: a value computed in an expression is outside every textual guard. `preset_max_members` is extracted and tested directly for that reason — a literal there passes all six, and only diverges once the constant moves, which is what `preset_digest_caps_members_at_the_shared_constant` asserts.

`#[cfg(test)]` fixtures are exempt; the cut is the gated *module*, never a gated item, since `search/shared.rs` gates a single `pub fn` among production code.

## How to verify

```bash
cargo test --lib defaults
```

Every guard is mutation-verified — a literal `default_value_t = 10`, a local `default_min_size`, `(default: 200)` and `Default: 200` in a description, `max_members_per_type: 50` in `core.rs`, a literal `impl Default` behind a mid-file `#[cfg(test)]` item, and a generic or next-line-brace `impl Default`. Each turns exactly one test red and names the file, the line, and the text.

Behavior: MCP `tools/list` is byte-identical to `main`. `--help` differs in exactly nine rendered lines, where a doc comment restated a number clap already prints — each still carries its own `[default: …]`. As a cross-surface check, setting `MAX_MEMBERS` to 3 makes `ast-bro digest src` and MCP `tools/call digest` both emit 14 truncation markers; before this change the CLI emitted none.

Full suite: 26 test binaries green, `cargo clippy --all-targets` clean apart from a pre-existing `no_effect_replace` warning in `src/search/chunker.rs`.

## Review

Three rounds of cross-review (Codex plus an independent Claude reviewer). Twelve findings, all from the Claude side — Codex returned "no actionable regressions" in all three rounds, including the round where the CLI digest cap was provably broken. Two were major: the `--preset digest` literal, and the first version of the struct guard omitting `max_members_per_type` — the very field the class was found through, which is what retired the hand-written allowlist in favor of the two rules above.

## Related

Companion to #56, which records the CLI/MCP convention in `AGENTS.md`. Shared constants make a *value* impossible to diverge; neither this nor that note catches an argument that exists on only one surface, which is #58.


aeroxy:
@vlsi Reviewed at `f71c35d` against base `d993b62`, out-of-tree via `git archive` (another checkout was busy), so nothing here touched the working tree.

**No correctness defects — the value half of the invariant holds.** Verified rather than eyeballed:

- All ~45 conversion sites audited 1:1, then confirmed at runtime: `--help` across 14 subcommands yields exactly the pre-PR defaults (`.`/200/2/all, 8000, 1, 12, 3, 16, 10, 20), and an MCP `tools/list` round-trip renders all 36 schema descriptions byte-identically to the old hard-coded text. No constant is swapped.
- The six new guards pass, the build is warning-free, 409/409 lib tests green.
- `deps_e2e::go_duplicate_module_resolves_within_the_importers_copy` fails, but it fails identically on `d993b62` — pre-existing, not from this PR (looks like the #53 Go work).

Four follow-ups, all low severity. None blocks merge; every one is a gap in the *guards* rather than in behavior, which matters only because the guards are the point of the PR.

**1. `is_literal_field` recognizes ints and `PathBuf::from("…")` only** — `src/defaults.rs:215`. String and float defaults slip past all three struct / `impl Default` guards, so `IMPACT_MODE` — the module's one non-numeric constant — is exactly the blind spot. A literal `"all"` in a struct body passes all six guards.

**2. `impact --mode`'s doc comment hard-codes the default** — `src/lib.rs:488` says ``or `all` (default)`` while `:489` reads `IMPACT_MODE`. `schema_descriptions_interpolate_their_default` covers MCP descriptions only, so the clap `--help` surface can still drift into contradicting its own `[default: …]`. Nine doc comments were cleaned up here for restating a number; this one restates it as a word.

**3. `sources()` walks three files where its own hint cannot compile** — `src/defaults.rs:156` walks all of `src/`, which includes `src/main.rs`, `src/bin/ast-outline.rs`, and `src/bin/sb.rs`. Those are separate bin crates, and `mod defaults;` (`src/lib.rs:10`) is private, so neither `crate::defaults::X` nor `ast_bro::defaults::X` resolves there. Latent today — all three are thin shims with no clap defaults — but a default added to one would fail the guard with advice that will not build. `pub mod defaults;`, or skipping the bin targets in the walk, closes it.

**4. `ROOT` still has the expression-position hole that `MAX_MEMBERS` had closed** — `PathBuf::from(".")` at `src/lib.rs:1058`, `src/run/cli.rs:22`, `src/mcp/tools.rs:1087`. The PR names this residual class and extracts `preset_max_members` to close it for the member cap; `ROOT` is the same shape, and it is the one affected default with a live second surface. (`src/lib.rs:1615` is a cwd fallback, not a walk root — leave that one.)

Worth noting for scope: #58 states that this PR closes the value half of CLI/MCP drift. It does, for values that reach an attribute — 1, 3, and 4 above are the residue where a value does not, so they belong with #57's own class rather than with #58's argument-existence work.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #61, opened 2026-08-12
implements: report the hierarchy a scoped walk could not read

vlsi:
## Why

`implements <Type> <path>` returns a silently partial answer when the walk root does not contain the type's declaration and its intermediate classes. On [pgjdbc](https://github.com/pgjdbc/pgjdbc) at `9874eb8`:

```
sb implements BinaryCodec pgjdbc/src/main/java/org/postgresql/jdbc/codec/   -> 7 matches
sb implements BinaryCodec pgjdbc/src/main/java                             -> 52 matches, all 52 inside that same codec/ directory
```

Narrowing the walk to the directory that holds every answer loses 45 of 52, at exit 0, with nothing on stderr. Transitive resolution has to follow `extends AbstractTextCodec` / `extends StreamingBinaryCodec` / `extends PrimitiveBinaryDecoder`, all of which live outside `codec/`; when the walk cannot see them the chain stops and the classes below it are simply absent.

That contradicts the error contract in `wiki/architecture.md`: *"an empty answer has to say how much ground it covers, or a caller cannot tell an honest zero from a scope that collapsed."* The same reasoning applies to a partial answer. `implements` already separated "no such type" (exit 2) from "type exists, zero implementors" (exit 0); the missing third state is "type resolved, but the walk could not see part of its hierarchy".

## What

A transitive answer now reports the links it could not read, in the shape `frontier_truncated` established for the depth-bounded commands:

- `hierarchy_truncated` and `unseen_bases` (`{name, path}`) in `ast-bro.implements.v1`;
- a `# note:` on stderr in text mode, naming those bases and the files that declare them;
- the same lines in front of the MCP response body, which has one channel.

**What earns a place in that list is evidence**, the standard `path_repair` already holds a path hint to. A base is reported only once the project is found to declare it outside the walk *and* the declarations read prove that base inherits the target. Neither weaker rule survives contact with real code: the raw list of bases a walk fails to resolve runs to 70 names on a Java source tree, nearly all of them JDK types no wider path would resolve, and "everything not yet ruled out" is again almost every base in a language whose types bottom out in library interfaces — the note would then print the same thing whatever the target (measured: byte-identical notes for two different targets on spring-framework, and `Error`/`Exception`/`IOException` offered for `AbstractList` on the JDK). Proof needs the chain followed to its end, so the probe repeats until the declarations it parsed name nothing new; at that fixed point a name the project never declares belongs to a library, and a library type cannot inherit a type of this project.

The probe finds a declaration **by file name**, ignoring case and word separators (`StreamingBinaryCodec.java`, `streaming_binary_codec.py`, `Middle.d.ts`), across every parseable extension rather than only the ones the walk saw, and across every project root the walk reached into. That is one directory walk per root and a parse of the few files a name matched — no file contents are scanned. Cost on pgjdbc: **13 ms** over a 27 ms query. The first cut at this scanned the text of every project file for every unresolved name and cost 8.4 s on the JDK repository.

The price is a type whose file is not named after it — an inner type, a Go file holding several. The probe cannot place those, and since an unplaceable name also ends the descent proof above it, silence would be claiming coverage the search never had. Those names come back as `unlocated_bases`, readable in one direction: **empty** means every name was accounted for and `hierarchy_truncated` is exact; **non-empty** is the normal state of a Java answer and says nothing about that particular answer. An answer of zero also says so on stderr, where "no such subtype" and "the link is in a file I could not find" are the two readings a caller has to tell apart; it stays off non-empty answers, where a line firing on nearly every query is the failure mode `frontier_truncated` avoids by staying quiet at depth 1.

Two things are deliberately **not** gaps. The target's own declaration sitting outside the walk is not one — `find_implementations` seeds direct matches from base-name equality and never reads it, so the ordinary api/implementation split is already complete. And a base written qualified (`third_party.Middle`) is matched by its simple name, because that is what the resolver being predicted does: on a wider walk `find_implementations` reports the collision as a match, so a stricter probe would contradict the answer it qualifies. `--direct` is never qualified at all.

`src/hierarchy.rs` is new; `wiki/architecture.md` and the README CLI contract describe the rules above.

## How to verify

```bash
cargo test --test implements_hierarchy
cargo test --test mcp_e2e
```

21 tests in `tests/implements_hierarchy.rs` plus one MCP end-to-end test. Each covers an axis rather than the one reported input: the two-hop chain whose middle link is named only by an outside declaration, a base whose own ancestry ends in a library type, a project class colliding by simple name with the library type actually extended, a glob path argument, a cross-language walk, a multi-repository walk, `.d.ts`, snake_case file names, two files declaring the same simple name, and four arrangements of working directory versus project root. Every fix was checked by reverting it and confirming the matching test fails.

On the reproduction above the scoped walk still answers `7 match(es)` on stdout at exit 0, and stderr now carries:

```
# note: 6 base type(s) named by the walked files are declared outside the walked path(s). Any subtype that reaches 'BinaryCodec' through one of them is absent from this answer:
#   PrimitiveBinaryDecoder  pgjdbc/src/main/java/org/postgresql/api/codec/PrimitiveBinaryDecoder.java
#   ...
# add those paths to the walk to resolve the rest
```

The wide walk stays silent (0 bytes on stderr), which is the other half of the contract.

## Notes for review

- No issue number is cited in the code: this was reported outside the tracker, and `#35` is taken by an unrelated closed issue. Happy to file it and add the citations if you prefer the usual convention.
- A follow-up issue covers the limitation the file-name lookup leaves open (a link declared in a file not named after it, including a nested type).


aeroxy:
@vlsi — this is great work, thank you! 🙏

The PR description is a masterclass on its own. The part I appreciate most is that you didn't just pick a rule, you showed why the two weaker rules collapse on real code — "70 names, nearly all JDK types" and "byte-identical notes for two different targets on spring-framework" is exactly the kind of evidence that settles a design argument. The fixed-point probe in `src/hierarchy.rs` is a lovely piece of reasoning: at the fixed point, a name the project never declares belongs to a library, and a library type can't inherit a project type. That's a genuinely tight proof for something that could easily have been a pile of heuristics. And the `unseen_bases` / `unlocated_bases` split — one readable in a single direction — is the honest shape.

### What I checked locally

- Full `cargo test` on the branch: green, including your 21 new tests and the MCP e2e case.
- Your cost claim holds. On a synthetic 6k-file Java tree, a whole-project `implements` goes 0.95s → 1.01s; scoped queries are unchanged. On a ~2k-file TypeScript repo, 0.30s → 0.35s. The directory-walk-plus-targeted-parse design is doing what you said it does.
- The note itself reads well when it fires: it names the right file, and it spells paths the way the walk was spelled, so "add those paths to the walk" is literally copy-pasteable.

### 1. The probe never runs in the case that hurts most

`implements_target_exists` gates on the target's **declaration** being inside the walk, and it runs above the new gap code. So a scoped walk whose subtypes reach the target only through an outside link exits 2 before the probe ever gets a chance:

```
repro/api/Base.java       public interface Base {}
repro/api/Middle.java     public interface Middle extends Base {}
repro/codec/LeafOne.java  public class LeafOne implements Middle {}
repro/codec/LeafTwo.java  public class LeafTwo implements Middle {}
```

```console
$ sb implements Base .
# 3 match(es) for 'Base' (incl. transitive):
./api/Middle.java:1  interface Middle
./codec/LeafOne.java:1  class LeafOne [via Middle]
./codec/LeafTwo.java:1  class LeafTwo [via Middle]
# exit 0

$ sb implements Base codec
error: no type named 'Base' in the given path(s)
  A 0-match answer is only reported for types that exist. Check the spelling or widen the search path.
# exit 2
```

This is the same failure your PR is about — narrowing the walk to the directory that holds every answer loses every answer — just with 0 direct matches instead of pgjdbc's 7. And here the message is actively wrong: `Base` does exist, one directory over, and `unseen_bases` is holding the file that says so while the generic "widen the search path" hint gets printed instead. The MCP surface does the same thing (`isError: true`, same text).

It also nips at the rationale in the PR body. "The target's own declaration sitting outside the walk is not a gap — the ordinary api/implementation split is already complete" is true only when at least one subtype implements the target *directly*. Once that's zero, the same layout is rejected outright. `the_targets_own_declaration_is_not_a_gap` has that direct match in the fixture, so the 0-match variant isn't covered by the suite.

I don't think this is a regression you introduced — `main` behaves the same. But it's inside the scope you set, and it's the case where the note would be worth the most: exit 2 with "no such type" is the one answer a caller can't tell from a real typo, and you already have the evidence to say "it's in `api/`, add that path." Would you be up for extending the probe to run before that guard when the walk resolved zero matches, so the exit-2 hint can name a path? Happy for it to be a follow-up PR if you'd rather keep this one focused — just want us to decide deliberately rather than leave it.

### 2. This conflicts with #63, and the conflict isn't only textual

`git merge-tree` reports conflicts in `src/core.rs`, `src/lib.rs`, and `src/mcp/tools.rs` against #63. That one moves `implements` out into `src/implements.rs`, which is the easy half. The harder half is that #63 also changes what lands in `Declaration::bases` ("record the supertypes every language spells out") — and `bases` is the direct input to `unresolved_base_names`. So rebasing onto #63 will change what this note *says*, not just where the code lives.

My inclination is to land #63 first and rebase this on top, so your fixture assertions are written against the final base extraction rather than the current one. Shout if you'd rather go the other way and I'll rebase #63 instead.

### 3. Minor: `probed` is keyed by `stem_key`

In `hierarchy::unseen_bases`, the `probed` set is keyed by `stem_key(name)`, so two unresolved names that differ only in case or separators collapse to one entry — Go's unexported/exported `foo` / `Foo` pair is the realistic instance. Only the first reaches `asked`, so the second can never appear in `unlocated`.

`unseen` is unaffected, since `classify_unseen_bases` receives the full `unresolved` list and the index lookup parses both files anyway. It's narrow, but it does dent the "empty `unlocated` means `hierarchy_truncated` is exact" reading that the docs (rightly) lean on. Keying `probed` by the normalized name and keeping `stem_key` purely for the index lookup would close it.

### Two small things, not blockers

- **MCP note channel.** Prepending to the response body is right for today's single-channel contract. As we structure MCP returns more strictly I'd like the note to become a field rather than a prefix — noting it here so it's on the record, not asking you to change anything.
- **Issue citation.** Don't bother filing one. The PR description is a better source of truth than any issue I'd write, and I'll link this PR in the release notes.

Nothing above touches the core design, which I think is right. Get me an answer on (1) and a call on the (2) merge order and I'll approve. Really nice contribution. 🚀

PR #63, opened 2026-08-14
fix(implements): resolve a supertype reference to one declaration

vlsi:
### Why

`implements <Type>` matched a supertype reference against the target by simple name, so two types sharing one short name in different packages were treated as one and every subtype of either was reported as a subtype of both.

On pgjdbc, `implements BinaryCodec pgjdbc/src/main/java` returned 52 types. Three of them — `PrimitiveTextDecoder`, `PrimitiveTextEncoder`, `StreamingTextCodec` — extend `org.postgresql.api.codec.TextCodec`, an interface that never reaches `BinaryCodec`. They were dragged in because the unrelated final class `org.postgresql.jdbc.codec.TextCodec` is in that closure. The correct answer is 49.

This is not language-specific: the same collision exists wherever two modules can declare a type with the same short name. Reproduced before the fix in Java, C#, Kotlin, Python and TypeScript.

### What

The resolver moves out of `core.rs` into [`src/implements.rs`](../blob/implements-resolve-supertype-by-declaration/src/implements.rs), because it needs import information and the only import extractor in the project lives under `deps` — a dependency `core` must not take. `show.rs` and `impact.rs` are the precedent for a subsystem of its own.

A supertype reference now resolves in priority order:

1. a rename the referring file introduced (`import a.b.Base as Alias`), bounded by its own scope — a `use … as …` inside one Rust `mod` is not in scope in its sibling;
2. a qualifier, which rules a declaration *out* only where the language gives it a namespace to be compared with, since in Python, TypeScript and Rust `api.Base` is as likely a module alias as a package path. A leading segment some import binds is expanded first, whether it renames the type or names an enclosing one;
3. then narrowings over the declarations carrying the simple name — the referring file, a single-type import, the file's own package, an on-demand import.

Each narrowing applies only when it leaves a candidate standing, so a name that is unique in the walk resolves exactly as it did before and the rules can only change an answer that was a guess. An edge never crosses a language.

What no rule pins is dropped rather than guessed, and **reported**: an `ambiguous` array in the JSON, a `# note:` on stderr, the same line on the MCP body — the treatment truncation and depth cutoffs already get. The target goes through the same resolution, so `implements jdbc.codec.TextCodec` answers for one declaration, a bare `implements TextCodec` answers for both and says so, and a qualifier that fits nothing is a rejected query rather than an answer for everything. A target the walk does not declare is the ordinary case — `implements Runnable` over an application tree means the JDK's — and the references spelling that name are the answer.

Enums and C# delegates are collected as the declared types they are, so a qualified enum target answers 0 matches instead of exiting 2, and an enum implementing an interface appears in that interface's closure.

A separate commit fixes the adapters, because a resolver is worth nothing over declarations that record no supertypes — and five of them recorded none. Each asked the grammar for the wrong thing and got nothing back, while the signature still showed the clause, so `map` output looked right and the loss stayed invisible:

- Kotlin looked for a `delegation_specifiers` container the grammar does not emit;
- PHP and C++ read a `field()` those grammars expose as an ordinary child node;
- JavaScript was descended one level too far, having no `extends_clause` wrapper — `ns.Root` came out as `ns` and `Root`, a bare `Root` as nothing;
- Ruby kept the `<` operator in the name.

Two more surfaced once the fixtures carried a clause harder than `: Root`: a generic Rust `impl<T> Root for Hard<T>` never regrouped onto `Hard`, and TypeScript recorded a type argument as a base of its own. This changes `map` / `digest` / `show` / `surface` output for PHP, C++ and Ruby, whose signatures are built from `bases`; that is the point of the fix rather than a side effect, and it is pinned by a test.

A third commit fixes `deps::extract`, which sliced a Kotlin import's node text where the grammar folds a following comment into that node, so any Kotlin file whose last import is followed by a KDoc block lost that edge from the dependency graph.

### How to verify

```
cargo test --release
```

The new suites are `tests/implements_collision.rs` (31 tests) over `tests/fixtures/implements_collision/`, which declares one name in two packages per language, and `tests/adapter_supertypes.rs`, which walks a fixture per language and checks the list against the directory so a fixture added without a line fails instead of going unrun. Plus `deps_e2e::kotlin_import_survives_a_following_comment`.

Each of the four commits builds and passes the whole suite on its own.

Against pgjdbc at `9874eb876`, `implements BinaryCodec pgjdbc/src/main/java` now returns 49 types, set-identical to a package-aware declaration scan written without ast-bro. Timing is unchanged: 0.11s over 1.4k Java files.

Every fix carries a revert check — removing it fails the test written for it — and the fixtures range over the axes rather than the one reported input: target inside and outside the walk, bare and qualified spellings, agreeing and disagreeing qualifiers, renames wrapped and visibility-qualified, and 9/10/11 dropped edges against the note's display cap.

The nested-reference forms are the ones openrewrite/rewrite#8238 enumerated for the same problem in its Kotlin type attribution: fully qualified, relative to an imported outer type, through a rename, and relative to an imported nested type whose leaf repeats deeper. All four resolve; `tests/fixtures/implements_collision/nested/` covers them.

Known gaps, recorded in [wiki/architecture.md](../blob/implements-resolve-supertype-by-declaration/wiki/architecture.md) and out of scope here: Rust records `impl Trait for Type` but not a supertrait bound; Ruby records `<` but not an `include`d mixin; PHP enums are not extracted as declarations at all, so an `enum … implements` clause is invisible; a Scala `import` inside a class body is not extracted, so a supertype it would have pinned is reported as ambiguous; and the TypeScript extractor records no named binding for `import { Base as Alias }`, so a nested reference through such a rename stays reported rather than resolved.


aeroxy:
Hey @vlsi!

Amazing work! This PR is a masterclass in fixing silent, semantic bugs that plague LLM agents. The pgjdbc `TextCodec` collision is exactly the kind of issue that makes a coding agent hallucinate inheritance hierarchies. (And huge props to Claude Opus 5 for the pair programming! 🤖)

Your breakdown of the adapter fixes (Commit 2) is fascinating — it's wild that five languages were recording *no* supertypes at all while `map` output looked completely normal. Pinning them with per-language fixtures in `tests/adapter_supertypes.rs` is exactly how we keep a language from "going quiet" on us again. And catching the Kotlin `deps::extract` bug where a trailing KDoc block swallowed the import node is a brilliant catch.

Moving the resolver out of `core.rs` into `src/implements.rs` is the right call architecturally — `core` shouldn't take a dependency on `deps` just for import extraction. The priority order (Rename → Qualifier → Narrowing) is elegant, and dropping an edge rather than guessing is a big win for our error contract.

I verified the C++ base-clause kind filter across eight heritage shapes, the TS/JS `class_heritage` handling (including `type_arguments` and mixin calls), the Rust generic-impl regroup, the Go embedded-interface edges, and the MCP/CLI note plumbing — all correct. `cargo test --release` is green.

That said, I hit three things while poking at it that I think need to land before we merge. Fixtures below all reproduce against the release binary on this branch.

### 1. Namespace separators drop the edge entirely (blocking)

`_type_ref_segments` splits a reference on `::`, `.` **and** `\`, but `_qualifiers_agree` splits the *declaration's* qname on `.` only. `_collect_types_in` stores the raw namespace name, so PHP's `App\Base` and a C++17 `namespace A::B` each stay one segment on the declaration side and can never agree with a three-segment reference. Rule 1 then empties `live` → `BaseRef::None` and the edge disappears — not even reported in `ambiguous`, so it's silent.

```php
// base.php
namespace App\Base;
class Root {}

// child.php
namespace App\Models;
class Child extends \App\Base\Root {}
```
`implements Root` → **0 matches**. Change the namespace to a single segment (`namespace Base;` / `extends \Base\Root`) and it's 1.

Same in C++:
```cpp
namespace A::B { class Root {}; }              // → 0 matches
namespace A { namespace B { class Root {}; } } // → 1 match
```

This hits essentially every PSR-4 PHP project, which is most of them. I think the fix is to normalise the separator when building `qname` (or split on the same set in `_qualifiers_agree` / `_qualifier_matches`) so both sides agree on what a segment is.

### 2. A rename whose target is outside the walk falls through to the wrong type (blocking)

`_resolve_rename` returns `None` when a rename import binds the name but the target isn't in the walk, so the alias falls through to narrowing and lands on an unrelated in-walk type with the same simple name:

```kotlin
// api/Widget.kt
package api
open class Widget

// app/Child.kt
package app
import com.external.Thing as Widget
class Child : Widget()
```
`implements Widget` → reports `Child` as implementing `api.Widget`. Add `com.external.Thing` to the walk and it correctly flips to 0.

This is the merge the PR exists to prevent, and it's wrong in the direction that shows up in practice — an external dependency is normally *not* in the walk. It also contradicts the wiki line this PR adds ("a rename … shadows any declaration of that name elsewhere"). A rename binding the name should suppress the candidate regardless of whether the target resolved — unresolved-but-bound is still bound.

### 3. Ruby `Struct.new` superclasses now render as no superclass

The new extraction keeps only `constant` / `scope_resolution`, so:
```ruby
class Config < Struct.new(:a, :b)
end
```
now maps as bare `class Config`, positively asserting no parent. Low severity — the old output was malformed — but `Struct.new`, `Data.define` and `Class.new` are common enough that silently dropping them feels worse than rendering the call expression. Happy to treat this as a follow-up if you'd rather keep the PR tight.

### Non-blocking

- **Import re-reading on heavy collisions.** Since we only hit `deps::extract` on a name that actually collides, the overhead is minimal in the common case — but on a monorepo with lots of overlapping `Base`/`Config` types we'd re-parse the same file's imports repeatedly within one walk. Worth a `HashMap<PathBuf, Vec<RawImport>>` in the resolver context, or is the page cache cheap enough that it isn't worth the memory?
- **Schema version.** Did we mean to bump `ast-bro.implements.v1` → `v2`? Adding `ambiguous` is additive, but the resolution rules change the *shape* of the answer (cases that were exit 2 now resolve to 0), which downstream agents may key on.
- **The docstrings.** Please keep these exactly as they are — documenting JLS 6.4.1 for single-type imports shadowing same-named package types is invaluable for whoever next tries to "simplify" the resolver. The C# note about the enclosing namespace winning over using aliases is the kind of thing that gets refactored away and quietly breaks a language. ;)

Sorry to hold it up when the bulk of this is so clearly right — the adapter work and the architecture move I'd take as-is. Get 1 and 2 in with fixtures and I'll approve straight away. Fantastic contribution! 🚀

vlsi:
Both blockers are fixed, and Ruby came with them. Pushed as `b94b436` — 15 commits on top of `04d0811`, no rebase, so your review anchors still resolve.

Each of the three turned out to have siblings in the branch, so the fixes are at the level of the class rather than the instance.

## 1. Namespace separators (`739e8b7`)

Fixed in the one place that builds a qualified name. `_join_dot` now splits the name it appends on the same separators a type reference is split on, so both sides of every comparison are spelled in one alphabet. That single edit reaches `_qualifiers_agree`, `_qualifier_matches`, `_import_names` and the same-package rule, which is why it is there and not in the four readers. It also drops the generic arguments a declared name may carry, which the name index already did to its keys — a latent disagreement between the two.

Your PHP fixture answers 1, and so does `namespace A::B` in C++. The fixture carries a second namespace declaring the same name, so it also pins the half that matters: `implements App\Other\Root` is still 0, in both spellings of the target.

## 2. A rename whose target is outside the walk (`5430958`)

Unresolved but bound is bound. `_resolve_rename` now distinguishes "no import binds this name" from "one binds it to something the walk does not hold", and the second suppresses.

Two siblings came with it. The head of a qualifier is the other shape a file rebinds a name in — `use external::stuff as api` makes `api::Handle` mean `external::stuff::Handle` — and it fell through the same way, so a lone unrelated `Handle` captured the reference. And suppressing an edge under the wrong name is only half an answer if the right name cannot find it either, so `_spelled_reference` expands a rebound head on the out-of-walk path too: `implements external.stuff.Handle` now finds the subtype that spells it `api::Handle`. The merged-spellings note reads the same helper, so two aliases pointing at different modules no longer collapse into one spelling.

## 3. Ruby `Struct.new` (`a37a2d2`)

Taken here rather than as a follow-up, because the same defect was in three adapters. Each read its supertype clause through a list of the node kinds it expected a type to be, so any other kind vanished — and a missing base does not read as "unparsed", it reads as a claim that the class has no parent. Ruby, C++ and PHP now name what to drop instead: the operator and commas are anonymous, `public`/`private` are `access_specifier` nodes, and a comment is the one named child that names nothing. `< Struct.new(:a, :b)` and `< Data.define(:x)` render as written and resolve to nothing, which is the answer an out-of-walk supertype already gets.

## Non-blocking

**Import re-reading.** Already cached — `ImportIndex` keys by file for the life of a query, so a monorepo re-parses each file at most once however many references consult it. 2040 Java files with 2000 colliding `Config` references resolve in 0.15 s.

The cost model I documented was wrong, though, and I had to correct it twice. Extraction is not gated on the collision: the rename rule outranks the declaration index, so it runs before the index is consulted, and the qualifier expander and the two import narrowings ask nothing at all — they extract whenever they run, which is a reference of more than one segment or a name two declarations carry. A Java tree therefore extracts too, on exactly the queries the resolver exists for. Measured: 1.3 s against a 0.3 s `map` over 550 Python files of ~7 KB. Keying the probe on the name the reference spells would cut that and is the wrong trade — a false positive costs one reparse, a false negative deletes an edge and reports nothing. Both the `ImportIndex` docstring and the wiki now say this.

**Schema version.** Left at `v1`, and I would rather you called it. `ambiguous`, `target_candidates` and `external_spellings` are additive, and no field changes shape or meaning; what changed is which answers are correct, which is a bug fix rather than a schema change. A consumer that keyed on the old wrong answers is broken either way, and a bump would tell it nothing it can act on. Say the word and I will bump it.

**The docstrings.** Untouched, JLS 6.4.1 and the C# note included. The C# one earned its keep this round — see below.

## What the cross-review changed beyond your three

Five rounds, two reviewers, sixteen findings. Every round found a defect in the round before it, several of them in my own fixes. The ones that matter to you:

- **A bare target dropped a root that inherits from a sibling root.** `implements Base` over `a.Base` and `b.Base extends a.Base` reported `Leaf` alone while the note beside it claimed to cover both declarations, and `implements a.Base` reported the very type the bare form dropped. On pgjdbc, where the review measured it, that is `implements LibPQFactory`: 0 → 1.
- **This PR regressed rendering, and I had missed it.** Nesting PHP's statement-form namespace put its name in front of every symbol `surface`, `digest` and `show` print, and those join with dots — so a PSR-4 file came out as `App\Lib.Service`, which is neither PHP nor a target this tool takes back. A namespace declaration now carries dots in its name and the source spelling in its signature: `map` prints `namespace App\Lib`, a qualified path is spelled one way, and `show` splits a qualified name into segments so the path `surface` prints resolves. That last part was broken for every C# namespace before this branch — `surface` printed `Example.Api.Codec` and `show` rejected it.
- **A `using` written inside a namespace body was invisible.** `_walk_csharp` recursed into `namespace_declaration` but not into the `declaration_list` that holds its statements, so an alias in the body never reached the resolver: it lost its edge outright, or an outer binding of the same name won unopposed. That is a `deps` fix as much as an `implements` one.
- **`global::` lost the edge.** Read as a namespace segment it agreed with no declaration, so the qualifier rule rejected the very type the reference names — silently, since a rule that rejects every candidate leaves no ambiguity to report. Generated C# writes it on nearly every reference. `crate`, `super` and `self` were already anchors; `global` is the same kind of word, and the predicate now takes the language, since `global` is an ordinary directory name everywhere else.
- **Shadowing and ties.** An inner binding of a name shadows an outer one, whether or not either renames — that covers `using Root = A.Root`, the shape C# aliases exist for. Two bindings at the *same* scope tie, and a tie is not evidence: Python's `from a.mod import Base` followed by `from b.mod import Base` rebinds, so reading the first as decisive answered the losing side. Those now report the reference instead of picking.
- **A cycle is not a subtype relation.** `class A extends B` beside `class B extends A` had `implements A` reporting `class A [via B]`. Each branch of the walk carries the root it started from and may not arrive back at it, which keeps the sibling-root answer above.

I added one rule in round 4 and **withdrew it in round 5**, which is worth stating plainly. It made C# prefer a type in an enclosing namespace over a `using` alias — the positive half of the rule `own_scope_first` only does the negative half of. C# interleaves per level rather than ranking all members above all directives: for each enclosing namespace it consults the types declared there, then that namespace body's own alias and using directives, then moves outward. So the rule fixed the arm where the competing `using` is file-scope and broke the arm where it sits inside the namespace body — the placement this branch's own fixtures use, and the idiom. Half a language rule is worse than the gap it closes. The gap is written down in the wiki instead, including why the obvious fix is the wrong one.

## Three I deferred, all yours to call

- **A qualifier-rejected reference is dropped with no note.** Once the target resolves in the walk, the branch that matches out-of-walk references never runs, so a reference whose qualifier names a different type leaves without a word. The review measured it on pgjdbc: `implements Driver` prints `0 match(es)` and nothing on stderr, although `org/postgresql/Driver.java` reads `implements java.sql.Driver`. Dropping the edge is right — it is a different type, and `implements java.sql.Driver` finds it — but an agent that asked the bare name gets an unqualified zero. Adding a note class here touches the qualification surface #61 is about, so it belongs to whichever of these lands second rather than to me unilaterally.
- **TypeScript `import * as api from './base'`.** In-walk it resolves, including the two-declaration collision, through the on-demand rule. Out of walk it does not: `implements extlib.Root` misses `class Child extends api.Root`, because `deps::extract` keeps no local name for a namespace import. I wrote the resolver half, measured that it is dead without the extractor change, and reverted it — `deps` prints `[as X]` whenever `local_name` is set, so teaching the extractor changes `deps` output for every TypeScript namespace import. That is your call, not a resolver fix. Recorded in the wiki with its sibling: the brace form `import { Root as R }` records the module but not the item.
- **The C# lookup rule above**, if you want the per-level interleaving. `FileImport::scope_span` is the input it needs.

762 tests pass, clippy is clean apart from the pre-existing `chunker.rs:1558`, and every new test was checked against a reverted fix so none of them passes vacuously. On merge order: this is ready, and #61 rebasing on top is the way round I would do it, since `bases` settles here first.

aeroxy:
Hey @vlsi — thanks for turning these around so fast, and nicely done.

Both blockers are properly gone. Confirmed against your head (9eb9105):

- PHP `implements Root` across `App\Base` / `App\Models` → **1 match**, and the C++17 `namespace A::B` case agrees with the nested spelling now.
- The shadow rule reading off the *binding* rather than the rename is the right formulation — `import com.external.Thing as Widget` no longer falls through to an unrelated in-walk `Widget`. "Bound but unresolved is still bound" is exactly it.

I also re-checked the cycle handling, the multi-root per-branch visited set, the C# `declaration_list` recursion, and the PHP statement-namespace nesting — all sound.

A few things still outstanding, and **the second one is my fault, not yours** — you implemented exactly what I asked for and I didn't think through the blast radius.

### 1. C# same-file namespaces resolve backwards

Rule 2 (same file) is preempting rule 4 (same package), so a file with two namespaces gets it wrong in *both* directions:

```csharp
// one.cs
namespace A { public class Base {} }
namespace B { public class Child : Base {} }

// two.cs
namespace B { public class Base {} }
```

- `implements A.Base` → **1 match** (`Child`). False edge — C# namespaces don't leak sideways within a file, so `A.Base` is not in scope inside `namespace B`.
- `implements B.Base` → **0 matches**. The edge that genuinely exists is lost.

Same shape affects C++ and braced-namespace PHP. Being in the same file shouldn't outrank being in the same namespace, since the file is not the scope in any of these three languages.

### 2. `show` no longer accepts the namespace name `map` prints (my bad)

Normalising the separator into `name` fixed resolution but changed what the renderers emit, so the two commands now disagree about what a namespace is called:

```
$ ast-bro map ns.cpp
namespace A::B
    class Root  L1

$ ast-bro show ns.cpp 'A::B'
error: no symbol matching 'A::B' in 1 file(s) searched     # exit 2

$ ast-bro show ns.cpp 'A.B'
# ns.cpp:1-1 A.B (namespace)
namespace A::B { class Root {}; }                           # header and body disagree
```

Only the dotted form works, and it's a spelling the user never saw. PHP is rougher, because the backslash makes it look like a path problem:

```
$ ast-bro show base.php 'App\Base'
error: path not found: App\Base
  This argument reads as a path, not a symbol, so it was not searched for.
```

That sends the reader off hunting for a missing file when the symbol is right there. I should have asked for the normalisation to happen inside the resolver — a separate normalised field, or splitting on the same separator set in `_qualifiers_agree` / `_qualifier_matches` — and left the rendered `name` on the native spelling. Sorry for the round trip.

### 3. Aliased bare-relative import drops silently

`_resolve_relative` only strips `../` with a trailing slash, so a bare `..` survives as a literal path component and `_import_names` fails:

```python
# pkg/sub/child.py
from .. import Base as B      # → 0 matches
class Child(B): pass

from .. import Base           # → 1 match (control)
class Child(Base): pass
```

No note either, so it looks like a confident zero.

### Non-blocking

- `namespace A.B { class Child : Base }` doesn't prefer the enclosing `A.Base` — rule 4 uses exact package equality, but C# lookup walks outward through enclosing namespaces. Currently reports ambiguity and 0. This one at least says it's unsure rather than inventing an answer, so I'm happy to take it as a follow-up.
- An unrelated `import os as o` in a file makes the `ambiguous` note disappear for an edge that was already (correctly) dropped. The resolution is unaffected — I checked — so it's purely a reporting gap, but the note is the whole point of that section.

---

None of this is urgent. Enjoy the week off — hope it's something good, and pick this back up whenever you actually feel like it, not before. If you'd rather not carry #2 at all, say so and I'll do that one myself since I'm the one who sent you down that path.

Thanks again — the adapter work alone made this worth it. 🚀

